PR #4, opened 2020-06-24
save examples of request parameters

uiur:
same as #3 

also fixes an issue that request body changes: `{ name: 'foo' }` ->  `{ name: 'foo', table: { name: 'foo' } }`

### DIFF
--- a/lib/rspec/openapi/record_builder.rb
+++ b/lib/rspec/openapi/record_builder.rb
@@ -15,7 +15,7 @@ class << RSpec::OpenAPI::RecordBuilder = Object.new
       path: route.path.spec.to_s.delete_suffix('(.:format)'),
       path_params: request.path_parameters,
       query_params: request.query_parameters,
-      request_params: request.request_parameters,
+      request_params: raw_request_params(request),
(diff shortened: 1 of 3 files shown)


lib/rspec/openapi/record_builder.rb
```diff
@@ -35,4 +35,14 @@ def find_route(request)
     end
     raise "No route matched for #{request.request_method} #{request.path_info}"
   end
+
+  # workaround to get real request parameters
+  # because ActionController::ParamsWrapper overwrites request_parameters
+  def raw_request_params(request)
```
uiur, line 41: I've found a workaround to get real request params but it's not beautiful. Any idea?

k0kubun: I'm not sure if I understand the background. Could you add a request spec under spec/requests which fails without the workaround? Currently just using `request.request_parameters` generates the same doc.

k0kubun: Oh, I just forgot using `OPENAPI=1`. I'll take a look at what's going on this night.

PR #9, opened 2020-09-09
Avoid JSON::ParserError when response body is no content

uiur:
If an api endpoint returns response with no content like `head :ok`,
`rspec-openapi` raises json parse error.

```ruby
      Failures:

         1) Tables #destroy returns no content if specified
            Failure/Error: response_body: response.parsed_body,

            JSON::ParserError:
              767: unexpected token at ''
            # ./lib/rspec/openapi/record_builder.rb:37:in `build'
            # ./lib/rspec/openapi/hooks.rb:12:in `block in <top (required)>'
```

This PR handles this parse error. It renders openapi spec without response body in this case.


k0kubun (approved):
LGTM. Thanks!

PR #10, opened 2020-10-21
Support tag generation

natsuokawai:
This pull request is for generating tags.
openapi tags: https://swagger.io/docs/specification/grouping-operations-with-tags/

Tags group some paths into one, as follows.

スクリーンショット 2020-10-21 15 42 06


k0kubun (approved):
Great, thanks!

PR #11, opened 2020-11-13
Fix float type

natsuokawai:
This PR fixes float type error by implementing `build_format` method for `format` keyword.

When you load a schema file including 'type: float' in Swagger Editor, structural error occurs.

スクリーンショット 2020-11-13 12 10 40
スクリーンショット 2020-11-13 12 10 46

In OAS3, `float` is not allowed for `type` and should be written as follows:
```yml
size:
  type: number
  format: float
```
https://swagger.io/docs/specification/data-models/data-types/


k0kubun:
Hmm, interesting. Thanks!

PR #12, opened 2020-11-13
Support multipart/form-data

natsuokawai:
Currently multipart/form-data in ( request | response ) body is not supported.

スクリーンショット 2020-11-13 17 21 05

According to the document, multipart/form-data is described as `type: string, format: binary`.
https://swagger.io/docs/specification/describing-request-body/file-upload/

I'm not sure if this is the right way to determine multipart/form-data in the `build_type` method, is this implementation acceptable?

I also think the example needs to be handled well because there's a lot of unnecessary information.


k0kubun:
Hmm, your test runs successfully in my local environment, but CI is not happy about it for some reason.

k0kubun:
Ah, no, I just forgot switching the branch. I reproduced the failure.

> I'm not sure if this is the right way to determine multipart/form-data in the build_type method, is this implementation acceptable?
>
> I also think the example needs to be handled well because there's a lot of unnecessary information.

Agreed. Though it's probably better than crashing it with NotImplementedError. So let me merge this for now.

natsuokawai:
Thank you🙏

PR #16, opened 2021-01-12
Add application version field

jorbs:
Currently, the `version` field doesn't output to the generated YAML file and, as OpenAPI specification states, it is a required field. See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#fixed-fields-1.

Also, when we try to import the YAML into Postman, it throws `format not recognized` because there is no `version` field on the output file.

So this PR let's the user specify an application version to be set in the YAML file and use `1.0.0` as default.


k0kubun (approved):
Thanks!

PR #17, opened 2021-01-27
fix Multipart/form-data example object to filename

fruitriin:
OpenApi v3 parse `&1 !ruby/object:File {}` to unknown tag.
thus, sample openapi.doc also can't open swagger-ui, swagger-editor, so on.
スクリーンショット 2021-01-27 16 08 54

I fix put filename instead of ActionDispatch::Http::UploadedFile when create example.
I'm not sure to filename is enough params to example. you should fix as you like.
Is this implementation acceptable?


k0kubun (approved):
I think it is acceptable. Thank you!

PR #18, opened 2021-01-27
fix Type `null` to nullable: true

fruitriin:
when spec response null, convert to type: null
but OASv3 don't allow to type: null. Instead to use nullable: true.

Thus I fix build_type and add sample response.
Note: I added null to structure table, but I can't judge to better to create new spec.


fruitriin:
I'll fix confilcts

fruitriin:
wow, thanks

k0kubun:
No problem, thank you for your patch :) Apparently YAMLs are not generated consistently in different environments... Anyway, merging.

PR #22, opened 2021-02-28
Additional configuration options

ekampp:
This introduces configuration options for:

Info block:

* title
* description

Path item object:

* summary
* description

Path item object example:

* description (previously description_builder)
* headers

Other root objects:

* Tags


lib/rspec/openapi.rb
```diff
@@ -2,13 +2,29 @@
...
+      description
+      description_builder
+      enable_example
+      path
+      summary_builder
+      tags
+      title
+    )
+
+    attr_accessor(*ACCESSORS)
```
k0kubun, line 28: I prefer not to expose an unnecessary constant.

```suggestion
    attr_accessor(
      :application_version,
      :comment,
      :description,
      :description_builder,
      :enable_example,
      :path,
      :summary_builder,
      :tags,
      :title,
    )
```

k0kubun (approved):
The options look good, thanks!

k0kubun:
Could you fix the CI failure first though?

lib/rspec/openapi.rb
ekampp: The cool thing about the constant is that it can be tested, and it's easier to keep alphabetically sorted. But I will change that before submitting it for final approval 👍🏼

ekampp:
@k0kubun, I'm still working on this. I will, of course, make sure that there are (a) specs for my changes and (b) other specs pass.

ekampp:
@k0kubun. I'm about done with this. Can I bother you for a review and help me understand where the source YAML file that the failing spec is comparing to is located to update it? 

Thanks.

k0kubun:
> Can I bother you for a review and help me understand where the source YAML file that the failing spec is comparing to is located to update it?

You can just apply the diff shown in the CI output, which also has the location of the broken YAML files.

ekampp:
@k0kubun, I have updated the generated YAML files as well as committed the generated JSON files. 

I don't have any failing specs locally, but the automated tests are still failing. It looks like there are a few problems. 

1. Some of them maybe because I sorted the individual examples in the paths alphabetically, 
2. Some of them look like whitespace diff differences related to `null_sample`. Have you experienced this before?

k0kubun:
Yes, I've experienced it. YAML's behavior for null seems different across environments. I think we could test null differently to fix it, but in the mean time, please just apply the diff shown in *the CI output* as I said.

ekampp:
Hi, @k0kubun,

> but in the meantime, please just apply the diff shown in the CI output as I said.

I already did that locally. And I don't have any errors locally with the committed diffs. So I'm unclear where the problem comes from.

k0kubun:
I'm sorry that I gave you a confusing explanation, but what I wanted to mean by CI doesn't include your local environment, because I don't believe your local environment continuously integrates this repository.

Let me rephrase: I expect you to copy https://github.com/k0kubun/rspec-openapi/runs/2025523636 instead of whatever output you have in your local machine.

ekampp:
Hey, @k0kubun. Sorry for not having done much here. It's been a busy few weeks. 

From the diff in the automated tests, it's hard for me to copy the output. So, I'm sort of stuck here. Do you want me to improve the specs instead of _just copying it_ from the output?

k0kubun:
> Do you want me to improve the specs instead of just copying them from the output?

If you know and can tell me why a generated YAML is different between your environment and CI, I can improve the spec for you. But otherwise, I expect you to just make the CI successful by modifying YAML files under `spec`.

ekampp:
> why a generated YAML is different between your environment and CI

If I knew that, I would already have fixed the issue 😅

Can I bother you to try and run the specs locally and see if anything springs out at you?

k0kubun:
I'm sorry if it wasn't clear, but the YAMLs behave differently on my machine too. So we're under the same situation.

k0kubun:
You could split your pull requests into smaller ones for each feature if it's hard to fix too many failures at once.

ekampp:
If there is a consistent inconsistency in tests, doesn't that indicate that there may be some flakiness around the tests that we can possibly fix?

k0kubun:
You could say it's flaky, but the behavior and the failure are not random. You can make the CI green for sure.

ekampp:
I can. The current setup _could_ produce inconsistent output for developers?

k0kubun:
Yup. I'm sorry but I'll take a look at it when I can, but in the meantime I want you to just pass the CI if you want this PR to be merged.

ekampp:
I have some time this week. I will have another look at it.

k0kubun:
Got it. Please take your time, and mention me when it becomes ready to be merged.

k0kubun:
Now we also have conflicts. Let me close this until it becomes actionable on my end. Thanks.

PR #23, opened 2021-04-24
RSpec converts YAML files to Hash before comparison

kyoshidajp:
Hello, thank you for your nice tool.

I noticed the following failed test.
https://github.com/k0kubun/rspec-openapi/pull/22/checks?check_run_id=2051943034

I was able to reproduce it in my environment, too. The comparison of YAML files seems to be a bit buggy.

My idea is to convert the YAML files to Hash before the comparison.

### DIFF
--- a/spec/rspec/rack_test_spec.rb
+++ b/spec/rspec/rack_test_spec.rb
@@ -1,5 +1,6 @@
 require 'spec_helper'
 require 'fileutils'
+require 'yaml'
 
 RSpec.describe 'rack-test spec' do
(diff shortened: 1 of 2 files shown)


k0kubun:
Interesting idea. One of the benefits behind the original implementation was to be able to see what YAML changes are made by the current code locally, but given that we could see diffs on GitHub PRs and updating the YAML is not difficult without the diff either, this idea may be better. I might change my mind later, but let's try this for now.

PR #24, opened 2021-04-24
Fix error of response example when Content-Disposition header is inline

kyoshidajp:
# The Issue

My action returns binary (Image) data using send_data with `disposition: "inline"` like this.

```ruby
class ImagesController < ActionController::Base
  protect_from_forgery

  def show
    path = "/file/to/image.png"
    image = File.open(path, "r")
    send_data image.read, type: "image/png", disposition: "inline"
  end
end
```

Generation was failed.

```ruby
require "rails_helper"

RSpec.describe "Images", type: :request do
  describe "show" do
    it "return 200" do
      get "/image/:id" do
        expect(response.status).to eq 200
      end
    end
  end
end
```

```
% export OEPNAPI=1
% bundle exec rspec
...
An error occurred in an `after(:suite)` hook.
Failure/Error: elsif o =~ /\n(?!\Z)/  # match \n except blank line at the end of string

ArgumentError:
  invalid byte sequence in UTF-8
# /file/to/rspec-openapi/lib/rspec/openapi/schema_file.rb:29:in `write'
# /file/to/rspec-openapi/lib/rspec/openapi/schema_file.rb:15:in `edit'
# /file/to/rspec-openapi/lib/rspec/openapi/hooks.rb:19:in `block in <main>'
```

# Solution

If `Content-Disposition` response header is presented, then `example` will not be shown.

# Other information

This PR's spec will fail before merging #23 or other fixing PRs to be modified.

### DIFF
--- a/lib/rspec/openapi/record.rb
+++ b/lib/rspec/openapi/record.rb
@@ -11,5 +11,6 @@ RSpec::OpenAPI::Record = Struct.new(
   :status,                # @param [Integer] - 200
   :response_body,         # @param [Object]  - {"status" => "ok"}
   :response_content_type, # @param [String]  - "application/json"
+  :response_content_disposition, # @param [String]  - "inline"
   keyword_init: true,
 )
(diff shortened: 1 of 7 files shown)


k0kubun:
Is it possible to add a spec for this?

kyoshidajp:
I forgot to add spec. I added it.

k0kubun (approved):
Thanks!

PR #26, opened 2021-06-11
Rails 6.1 deprecates content_type in favor of media_type

bf4:
In Rails 6.0 there's a lot of "DEPRECATION WARNING: Rails 6.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead. "


k0kubun (approved):
Thanks!

PR #27, opened 2021-06-11
Handle schema errors without failing the build

bf4:
Feature: Don't crash the build :) Skip bad records.

One of our requests raises "NotImplementedError: type detection is not implemented for: #<JSONAPI::Exceptions::BadRequest: JSONAPI::Exceptions::BadRequest>>" which causes the whole build to fail.

The specific error is " request_params={"_parser_exception"=>#<JSONAPI::Exceptions::BadRequest: JSONAPI::Exceptions::BadRequest>},"

I was thinking it would be nice to use objects here instead of block scope, but that's for another day.

For example, then I could more easily write my own hook logic that didn't rely on the `example.metadata[:type] == :request`

### DIFF
--- a/lib/rspec/openapi/hooks.rb
+++ b/lib/rspec/openapi/hooks.rb
@@ -6,6 +6,7 @@ require 'rspec/openapi/schema_file'
 require 'rspec/openapi/schema_merger'
 
 records = []
+records_errors = []
 
 RSpec.configuration.after(:each) do |example|
   if example.metadata[:type] == :request && example.metadata[:openapi] != false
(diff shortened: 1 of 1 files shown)


lib/rspec/openapi/hooks.rb
```diff
@@ -19,7 +20,21 @@
...
+        RSpec::OpenAPI::SchemaMerger.reverse_merge!(spec, RSpec::OpenAPI::SchemaBuilder.build(record))
+      rescue => e # e.g. SchemaBuilder raises a NotImplementedError
+        # NOTE: Don't fail the build
+        records_errors << [e, record]
+      end
     end
   end
+  if records_errors.any?
+    error_message = <<~EOS
+      RSpec::OpenAPIGot errors building #{records_errors.size} requests
```
k0kubun, line 33:
```suggestion
      RSpec::OpenAPI got errors building #{records_errors.size} requests
```
?

k0kubun (approved):
Thanks :+1:

PR #28, opened 2021-06-14
FIX: NotImplementedError inherits from Exception (not StandardError)

bf4:
Sorry that I missed this distinction in the other PR.  I didn't thoroughly test it in my code :(

I think perhaps the SchemaBuilder should raise an rspec-openapi specific error, and have that inherit from StandardError. I'm happy to make that change instead of this one.


k0kubun (approved):
:ok_hand:

PR #29, opened 2021-06-16
Build deeply nested parameters

bf4:
Build deeply nested parameters

e.g. with query params { "filter" => { "name" => "Benjamin" } }

- now: `build_parameter_name("filter", { "name" => "Benjamin" })` #=> filter[name]
- before: `"filter".to_s` #=> filter

- [ ] failing test building out multiple parameters. (I'm not yet sure why it didn't rebuild to the yaml or test it)

xref https://github.com/OAI/OpenAPI-Specification/issues/1706 https://github.com/ahx/openapi_first/pull/75/files

### DIFF
--- a/lib/rspec/openapi/schema_builder.rb
+++ b/lib/rspec/openapi/schema_builder.rb
@@ -50,7 +50,7 @@ class << RSpec::OpenAPI::SchemaBuilder = Object.new
 
     record.path_params.each do |key, value|
       parameters << {
-        name: key.to_s,
+        name: build_parameter_name(key, value),
         in: 'path',
         required: true,
         schema: build_property(try_cast(value)),
@@ -60,7 +60,7 @@ class << RSpec::OpenAPI::SchemaBuilder = Object.new
 
     record.query_params.each do |key, value|
       parameters << {
-        name: key.to_s,
+        name: build_parameter_name(key, value),
         in: 'query',
         schema: build_property(try_cast(value)),
         example: (try_cast(value) if example_enabled?),
@@ -69,6 +69,16 @@ class << RSpec::OpenAPI::SchemaBuilder = Object.new
 
     return nil if parameters.empty?
     parameters
+  end
+
+  def build_parameter_name(key, value)
+    key = key.to_s
+    if value.is_a?(Hash) && (value_keys = value.keys).size == 1
+      value_key = value_keys.first
+      build_parameter_name("#{key}[#{value_key}]", value[value_key])
+    else
+      key
+    end
   end
 
   def build_request_body(record)
--- a/lib/rspec/openapi/schema_merger.rb
+++ b/lib/rspec/openapi/schema_merger.rb
@@ -26,10 +26,18 @@ class << RSpec::OpenAPI::SchemaMerger = Object.new
   # Should we probably force-merge `summary` regardless of manual modifications?
   def deep_reverse_merge!(base, spec)
     spec.each do |key, value|
-      if base[key].is_a?(Hash) && value.is_a?(Hash)
-        deep_reverse_merge!(base[key], value)
+      base_key_value = base[key]
+      if base_key_value.is_a?(Hash) && value.is_a?(Hash)
+        deep_reverse_merge!(base_key_value, value)
       elsif !base.key?(key)
         base[key] = value
+      elsif base_key_value.is_a?(Array) && value.is_a?(Array)
+        if key == "parameters"
+          # merge arrays
+          base[key] |= value
+        end
+      else
+        nil # no-op
       end
     end
     base
--- a/spec/rails/doc/openapi.yaml
+++ b/spec/rails/doc/openapi.yaml
@@ -24,6 +24,24 @@ paths:
         schema:
           type: integer
         example: 10
+      - name: filter[name]
+        in: query
+        schema:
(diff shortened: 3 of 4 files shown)


k0kubun:
> failing test building out multiple parameters. (I'm not yet sure why it didn't rebuild to the yaml or test it)

Yeah, I'm curious why the test is passing without any changes too.

bf4:
Since neither rspec nor rake ran all the specs, I ran `OPENAPI=1 bundle exec rspec $(git ls-files | grep '_spec\.rb' | xargs)` and it also generated a pretty big change to the roda spec and had one failure.

bf4:
I didn't really dig into how the merging works, but I confirmed on my local that it's still creating only one `parameters` element  even though the parameter name is now correct.

bf4:
I can also write a unit test in the meantime if you'd like. I tried to follow the pattern of the repo.

k0kubun:
> (❤️ for this project. I also liked the https://github.com/r7kamura/autodoc quite a bit. 🙌 )

🤝

spec/requests/rails_spec.rb
```diff
@@ -13,9 +13,21 @@
...
-      get '/tables', params: { page: '1', per: '10' }, headers: { authorization: 'k0kubun' }
-      expect(response.status).to eq(200)
+    context it 'returns a list of tables' do
+      it 'with flat query parameters' do
+        get '/tables', params: { page: '1', per: '10' }, headers: { authorization: 'k0kubun' }
+        expect(response.status).to eq(200)
+      end
+
+      it 'with deep query parameters' do
+        get '/tables', params: { filter: { "name" => "Example Table" } }, headers: { authorization: 'k0kubun' }
```
k0kubun, line 23: It's just a guess, but maybe the test is passing ignoring this because the above test generates parameters, and running this probably doesn't override it. So copying the same endpoint to somewhere else and calling it from this spec would probably approach the test issue.

bf4: True enough, though I'd like to be able to generate a spec like

```yaml
openapi: 3.0.3
info:
  title: rspec-openapi
  version: 1.0.0
paths:
  "/tables":
    get:
      summary: index
      tags:
      - Table
      parameters:
      - name: page
        in: query
        schema:
          type: integer
        example: 1
      - name: per
        in: query
        schema:
          type: integer
        example: 10
      - name: "filter[name]"
        in: query
        schema:
          type: string
        example: Example Table
      - name: "filter[price]"
        in: query
        schema:
          type: integer
        example: 0
      responses:
        '200':
          description: returns a list of tables
          content:
            application/json:
              schema:
                oneOf:
                  - type: array
                  items:
                    type: object
                    # etc
```

it's been a while since I studied the oas spec, so perhaps I'm misunderstanding it...

https://swagger.io/docs/specification/describing-responses/

```
Can I have different responses based on a request parameter? Such as:
GET /something -> {200, schema_1}
GET /something?foo=bar -> {200, schema_2}

In OpenAPI 3.0, you can use `oneOf` to specify alternate schemas for the response and document possible dependencies verbally in the response description. However, there is no way to link specific schemas to certain parameter combinations.
```

bf4: (If I remove the 'with flat query parameters' spec and rm spec/rails/doc/openapi.yaml and run `OPENAPI=1 bundle exec rspec spec/requests/rails_spec.rb` it generates a new spec but still with only one param. )

k0kubun: I agree. That's something we should fix as well. But if you're not trying to address that, maybe we should separate that for now and possibly merge them in a PR that address the problem.

lib/rspec/openapi/schema_builder.rb
```diff
@@ -71,6 +71,16 @@ def build_parameters(record)
     parameters
   end
 
+  def build_parameter_name(key, value)
+    key = key.to_s
+    if value.is_a?(Hash) && (value_keys = value.keys).size == 1
```
bf4, line 76: 1️⃣ build `{ "filter" => { "name" => "Benjamin" } }` as `filter[name]` instead of `filter`

lib/rspec/openapi/schema_merger.rb
```diff
@@ -26,10 +26,18 @@ def normalize_keys(spec)
...
-        deep_reverse_merge!(base[key], value)
+      base_key_value = base[key]
+      if base_key_value.is_a?(Hash) && value.is_a?(Hash)
+        deep_reverse_merge!(base_key_value, value)
       elsif !base.key?(key)
         base[key] = value
+      elsif base_key_value.is_a?(Array) && value.is_a?(Array)
+        if key == "parameters"
+          # merge arrays
+          base[key] |= value
```
bf4, line 37: 2️⃣ allow multiple parameter members to be built in the same request document

spec/rails/doc/openapi.yaml
```diff
@@ -151,6 +169,12 @@ paths:
         schema:
           type: integer
         example: 1
+      - name: id
+        in: path
+        required: true
+        schema:
+          type: integer
+        example: 2
```
bf4, line 177: bug when parameters are otherwise the same but have different examples

k0kubun (approved):
Thanks!

PR #32, opened 2021-11-22
New feature: enable example types configuration

justinschier:
## Current Behaviour
All OpenAPI requests get run on the `type: :request`.

## Desired Behaviour
We would like to be able to run on one or more arbitrary types, e.g. `type: :flow`. This would mean changing file 
[hooks.rb](https://github.com/k0kubun/rspec-openapi/blob/master/lib/rspec/openapi/hooks.rb) to be configurable.

This will allow us to generate our schema.yaml from our flow specs as well as from our request specs.

Thank you for this wonderful gem!

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,9 @@ EOS
 
 # Generate a custom description, given an RSpec example
 RSpec::OpenAPI.description_builder = -> (example) { example.description }
+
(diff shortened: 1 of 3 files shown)


k0kubun (approved):
:+1:

PR #36, opened 2022-03-15
Customizations via metadata

blocknotes:
Nice gem!

Can I propose you a small PR to allow more customizations via RSpec metadata?

Combined with the configuration:
`RSpec::OpenAPI.description_builder = -> (example) { example.metadata[:openapi_description] || example.description }`

A usage example could be:

```rb
  describe 'GET /api/v1/posts', {
    openapi_summary: 'list all posts',
    openapi_description: 'list all posts ordered by pub_date',
    openapi_tags: %w[v1 posts],
  } do
    # ...
  end
```

If you like the idea, I could update the specs and the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,22 @@ RSpec.describe '/resources', type: :request do
 end
 ```
 
+## Customizations
+
+Some examples' attributes can be overwritten via RSpec metadata options. Example:
+
+```rb
+  describe 'GET /api/v1/posts', openapi: {
+    summary: 'list all posts',
+    description: 'list all posts ordered by pub_date',
+    tags: %w[v1 posts],
+  } do
+    # ...
+  end
+```
+
+**NOTE**: `description` key will override also the one provided by `RSpec::OpenAPI.description_builder` method.
+
 ## Links
 
 Existing RSpec plugins which have OpenAPI integration:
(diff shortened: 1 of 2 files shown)


k0kubun:
Using metadata for the purpose sensible to me. Please do update the specs and update README to document it.

How about overloading `openapi:` option and nest keys under that though? i.e.

```rb
  describe 'GET /api/v1/posts', openapi: {
    summary: 'list all posts',
    description: 'list all posts ordered by pub_date',
    tags: %w[v1 posts],
  } do
    # ...
  end
```

When the option is taking a `Hash`, it's clear that it doesn't need to be `openapi: false`, so I think it doesn't conflict.

blocknotes:
> How about overloading `openapi:` option and nesting keys under that though? i.e.

Sounds good to me.
I was thinking the same at the beginning but I didn't want to replace your current behaviour.

About `description` - do you think that I could also add a customization there? Or it's better to avoid overriding the `RSpec::OpenAPI.description_builder` configuration method?

So this line: https://github.com/k0kubun/rspec-openapi/blob/master/lib/rspec/openapi/record_builder.rb#L53
Could become:

```rb
description: example.metadata.dig(:openapi, :description) || RSpec::OpenAPI.description_builder.call(example),
```

blocknotes:
On second thought, I think that it could be good to let override also the description.
So I proceeded in that way.

k0kubun (approved):
:+1:

PR #38, opened 2022-03-17
Handle Rails engine routes

### LINKED ISSUE #37: "No route matched for" error when using engine routes
Hey :)
In a project we have an engine which provides some routes.
When using _rspec-openapi_ the routes seem to be missing => `No route matched for GET /test` is raised.

I reproduced the issue here: https://github.com/blocknotes/rspec-openapi/tree/issue-engine-routes/issue

- `app.rb` is a one-file Rails app;
- `vendor/my_engine` is a sample engine with a test route;
- `spec` includes the request spec that I prepared.

You can try in the `issue` folder, running `OPENAPI=1 rspec` directly.

It could be similar to #25 - but there isn't an example there.

I'm also trying a fix here: https://github.com/blocknotes/rspec-openapi/commit/1201edd26bbe7f70c1752a0371337df7ebba940a

I saw that this line https://github.com/k0kubun/rspec-openapi/blob/master/lib/rspec/openapi/record_builder.rb#L79 recognize more than a route (it looks weird), but the first one is from the engine and let fail the method.

blocknotes:
Let iterate all the recognized routes including the ones coming from Rails engines.

As reference, the relevant part of `action_dispatch`'s `recognize`:
https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/routing/route_set.rb#L881-L892

Fixes #37

### DIFF
--- a/lib/rspec/openapi/record_builder.rb
+++ b/lib/rspec/openapi/record_builder.rb
@@ -19,8 +19,6 @@ class << RSpec::OpenAPI::RecordBuilder = Object.new
     # Generate `path` and `summary` in a framework-friendly manner when possible
     if rails?
       route = find_rails_route(request)
-      raise "No route matched for #{request.request_method} #{request.path_info}" unless route
-
       path = route.path.spec.to_s.delete_suffix('(.:format)')
       summary = route.requirements[:action] || "#{request.method} #{path}"
       tags = [route.requirements[:controller]&.classify].compact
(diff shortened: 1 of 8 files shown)


blocknotes:
Failing specs due to a route without a controller (block route in the engine).

```
            Failure/Error: tags = [route.requirements[:controller].classify]

            NoMethodError:
              undefined method `classify' for nil:NilClass
```

Should be fixed by #40

lib/rspec/openapi/record_builder.rb
```diff
@@ -78,13 +80,22 @@ def find_rails_route(request, app: Rails.application, fix_path: true)
       request.path_info = File.join(request.script_name, request.path_info)
     end
 
-    app.routes.router.recognize(request) do |route|
-      unless route.path.anchored
-        route = find_rails_route(request, app: route.app.app, fix_path: false)
+    found_route = nil
+    if app.respond_to? :routes
```
k0kubun, line 84: The tests don't fail even if you don't check `if app.respond_to? :routes`. Please remove this check or add a test that fails without this check.

k0kubun: addressed in https://github.com/k0kubun/rspec-openapi/pull/38/commits/81751cb19734dbe686d98aa357843be3e3c3acf1

k0kubun:
To mitigate the risk of this change, I removed changes that are not needed to pass the test case. https://github.com/k0kubun/rspec-openapi/pull/38/commits/81751cb19734dbe686d98aa357843be3e3c3acf1

k0kubun (approved):
Thank you!

PR #39, opened 2022-03-17
Fix parameters hash merging

blocknotes:
Hi :)
I think there's a typo / missing `|` in the parameters hash merging.

I get error when linting the yaml output:

```
Operations must have unique `name` + `in` parameters. Repeats of `in:path` + `name:id`
```

The relevant part of the test is this one:

```rb
  describe 'PATCH /api/v1/posts/:id' do
    context 'with valid data' do
      let(:post_params) { { title: 'Another title' } }

      context 'with a test post' do
        let(:test_post) { create(:post, title: 'Some title', description: 'Some description') }

        it 'updates the title' do
          expect {
            patch "/api/v1/posts/#{test_post.id}.json", params: { post: post_params }
          }.to(change { test_post.reload.title }.from('Some title').to('Another title'))
        end
      end

      context 'with a missing post' do
        before { patch "/api/v1/posts/0.json", params: { post: post_params } }

        it 'returns a not found error', :aggregate_failures do
          expect(response).not_to be_successful
          expect(response).to have_http_status(:not_found)
        end
      end
    end
  end
```


k0kubun:
Sorry, I can't remember the full context around this code by just looking at https://github.com/k0kubun/rspec-openapi/pull/29 and the code. Because the code is "merging" parameters, using `|` alone doesn't make me think it is a typo (because `|` merges arrays and `||` deson't).

Since you have a failing test in your description, could you just add a test case that fails without your implementation change?

blocknotes:
> Since you have a failing test in your description, could you just add a test case that fails without your implementation change?

Ok, anyway merging with only the pipe operator there looks odd to me.
In the example above, in the merging point:

```rb
[1] pry(#<Object>)> base[key]
=> [{"name"=>"id", "in"=>"path", "required"=>true, "schema"=>{"type"=>"integer"}, "example"=>0}]
[2] pry(#<Object>)> value
=> [{"name"=>"id", "in"=>"path", "required"=>true, "schema"=>{"type"=>"integer"}, "example"=>1}]
```

We could merge considering only `name` and `in` values considering the linting message produced by `@redocly/openapi-cli`.

blocknotes:
@k0kubun: I'm investigating more on this issue, here is the outcome:

In master branch, linting the current specs sample OpenAPI file with:
`npx @redocly/openapi-cli lint spec/rails/doc/openapi.yaml`

I get these errors:

```sh
No configurations were defined in extends -- using built in recommended configuration by default.

validating spec/rails/doc/openapi.yaml...
[1] spec/rails/doc/openapi.yaml:10:1 at #/servers

Servers must be a non-empty array.

 8 |   title: rspec-openapi
 9 |   version: 1.0.0
10 | servers: []
11 | paths:
12 |   "/tables":

Error was generated by the no-empty-servers rule.


[2] spec/rails/doc/openapi.yaml:179:9 at #/paths/~1tables~1{id}/get/parameters/1

Operations must have unique `name` + `in` parameters. Repeats of `in:path` + `name:id`.

177 |     type: integer
178 |   example: 1
179 | - name: id
180 |   in: path
  … |   < 3 more lines >
184 |   example: 2
185 | responses:
186 |   '200':

Error was generated by the operation-parameters-unique rule.
```

The first error is not relevant (it can be solved via `RSpec::OpenAPI.server_urls`).

---

While in my branch:

```sh
rm spec/rails/doc/openapi.yaml
OPENAPI=1 bundle exec rspec --order defined spec/requests/rails_spec.rb
npx @redocly/openapi-cli lint spec/rails/doc/openapi.yaml
```

The second error is gone.

What do you think?

k0kubun (approved):
It was much easier to understand the context for me now. Thank you!

PR #40, opened 2022-03-18
Fix some linter issues

blocknotes:
Hey :)

Checking with `@redocly/openapi-cli` linter I get some errors:
- `Expected type string but got null` related to the `tags` key when an empty value is exported;
- `Operation object summary must be non-empty string.` related to the `summary` key when an empty value is exported.

This PR is a proposal for minor improvements.

Please let me know your opinion.

### DIFF
--- a/lib/rspec/openapi/record_builder.rb
+++ b/lib/rspec/openapi/record_builder.rb
@@ -20,7 +20,7 @@ class << RSpec::OpenAPI::RecordBuilder = Object.new
     if rails?
       route = find_rails_route(request)
       path = route.path.spec.to_s.delete_suffix('(.:format)')
-      summary = "#{route.requirements[:action]}"
+      summary = route.requirements[:action] || "#{request.method} #{path}"
       tags = [route.requirements[:controller]&.classify].compact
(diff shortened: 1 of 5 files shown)


k0kubun:
Could we possibly add tests for this thing?

blocknotes:
> Could we possibly add tests for this thing?

Ok, if you run the spec only with the first commit (_Add a test with a block route_) you get this error:

```
# bundle exec rspec

         1) Extra routes #test_block returns the block content
            Failure/Error: tags = [route.requirements[:controller].classify]

            NoMethodError:
              undefined method `classify' for nil:NilClass
            # ./lib/rspec/openapi/record_builder.rb:24:in `build'
            # ./lib/rspec/openapi/hooks.rb:14:in `block in <top (required)>'
```

If you run the spec on the second commit (_Skip empty tags and fix routes without a Rails controller_) it will produce an OpenAPI YAML with an empty summary.

I removed the last commit (_When a route is not found print a warning in place of an exception_) to make some extra checks.

k0kubun (approved):

PR #44, opened 2022-03-26
Fix for engine routes (2)

blocknotes:
Hey @k0kubun.
I still have issues with an application with Rails engines... but it's tricky to create a spec for the case 😑 

That's what I see:

- with the current master version (`0.47`) I get again "No route matched for ..." exception;
- applying the PR change to `find_rails_route` (but keeping the raise line) it gets better, but it still fails when there's a test on a non-existing path;
- replacing the raise with the warn seems to solve the issue completely.

I think that the iteration of the routes with the current code still returns `nil` in some situations (probably the line `return find_rails_route(request, app: route.app.app, fix_path: false)`).

Think about it, if you don't feel confident to merge a PR without a good test it's ok (in any case having an alternative here could help someone else perhaps).

---

Relates to #38 and #37

### DIFF
--- a/lib/rspec/openapi/record_builder.rb
+++ b/lib/rspec/openapi/record_builder.rb
@@ -19,6 +19,8 @@ class << RSpec::OpenAPI::RecordBuilder = Object.new
     # Generate `path` and `summary` in a framework-friendly manner when possible
     if rails?
       route = find_rails_route(request)
+      return warn("RSpec::OpenAPI - No route matched for #{request.request_method} #{request.path_info}") unless route
+
       path = route.path.spec.to_s.delete_suffix('(.:format)')
       summary = route.requirements[:action] || "#{request.method} #{path}"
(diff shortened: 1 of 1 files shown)


k0kubun:
If you can't create a test case, please at least share a repository that reproduces it. You should at least have that one somewhere, and I think you can carve out what's still reproducing your issue to a small repository. Think about it, when I have no way to test my own code, it will randomly reproduce the same issue on your project again. So merging it temporarily doesn't mean anything if I can't reproduce it.

blocknotes:
> If you can't create a test case, please at least share a repository that reproduces it. You should at least have that one somewhere, and I think you can carve out what's still reproducing your issue to a small repository. Think about it, if I have no way to test my own code, it will randomly reproduce the same issue on your project again. So merging it temporarily doesn't mean anything if I can't reproduce it.

I tried to isolate the behaviour without success and the project is closed source.
It's fine anyway - I'll keep the commit in my fork.

PR #53, opened 2022-05-15
Remove TravisCI which seems no longer used

exoego:


k0kubun (approved):
didn't notice that. thanks :+1:

PR #55, opened 2022-05-16
Implement smarter merge

### LINKED ISSUE #48: More intelligent schema updates
https://github.com/k0kubun/rspec-openapi/blob/v0.4.8/lib/rspec/openapi/schema_merger.rb#L28

## Background
Currently, we have two ways to use rspec-openapi:

1. Delete the OpenAPI schema file and then run rspec-openapi for all tests to build it from scratch.
2. Just run rspec-openapi and update auto-generated fields while leaving all past changes.

By default, it works in mode 2 because it allows you to incrementally update the schema and also leave manual changes for fields that rspec-openapi cannot automatically generate.

## Problem
However, the default behavior, mode 2, sometimes has a problem. As described in https://github.com/k0kubun/rspec-openapi/issues/47, when you delete a parameter, even if you remove the parameter and rerun the same spec, it will not delete such obsoleted fields for you.

## Proposed Solution
We could possibly build each endpoint from scratch, but leaving keys that rspec-openapi does NOT support. When you run a single spec that requests a route, everything under the path/method should be erased while leaving keys that rspec-openapi cannot update as is, and then changes generated by rspec-openapi should be added to it.

It does mean that you will need to run multiple specs for the same endpoint when some tests are testing only some subset of parameters and you need to run all of them to document everything, but generally, running a single spec file or a describe block should be enough and it's probably not too hard.

exoego:
Closes #48 

This attempts to implement the behavior described in 
> everything under the path/method should be erased while leaving keys that rspec-openapi cannot update as is,
> and then changes generated by rspec-openapi should be added to it.

### DIFF
--- a/lib/rspec/openapi/schema_merger.rb
+++ b/lib/rspec/openapi/schema_merger.rb
@@ -25,22 +25,15 @@ class << RSpec::OpenAPI::SchemaMerger = Object.new
   # Also this needs to be aware of OpenAPI details unlike an ordinary deep_reverse_merge
   # because a Hash-like structure may be an array whose Hash elements have a key name.
   #
-  # TODO: Perform more intelligent merges like rerouting edits / merging types
   # TODO: Should we probably force-merge `summary` regardless of manual modifications?
   def deep_reverse_merge!(base, spec)
     spec.each do |key, value|
       if base[key].is_a?(Hash) && value.is_a?(Hash)
-        deep_reverse_merge!(base[key], value)
-      elsif !base.key?(key)
-        base[key] = value
-      elsif base[key].is_a?(Array) && value.is_a?(Array)
-        # parameters need to be merged as if `name` and `in` were the Hash keys.
-        if key == 'parameters'
-          base[key] |= value
-          base[key].uniq! { |param| param.slice('name', 'in') }
+        if !base[key].key?("$ref")
+          deep_reverse_merge!(base[key], value)
         end
       else
-        # no-op
+        base[key] = value
       end
     end
     base
--- a/lib/rspec/openapi/schema_merger.rb
+++ b/lib/rspec/openapi/schema_merger.rb
@@ -32,6 +32,14 @@ class << RSpec::OpenAPI::SchemaMerger = Object.new
         if !base[key].key?("$ref")
           deep_reverse_merge!(base[key], value)
         end
+      elsif base[key].is_a?(Array) && value.is_a?(Array)
+        # parameters need to be merged as if `name` and `in` were the Hash keys.
+        if key == 'parameters'
+          base[key] |= value
+          base[key].uniq! { |param| param.slice('name', 'in') }
+        else
+          base[key] = value
+        end
       else
         base[key] = value
       end
--- a/spec/rails/doc/openapi.json
+++ b/spec/rails/doc/openapi.json
@@ -106,6 +106,46 @@
         },
         "parameters": [
           {
+            "name": "page",
+            "in": "query",
+            "schema": {
+              "type": "integer"
+            },
+            "example": 1
+          },
+          {
+            "name": "per",
+            "in": "query",
+            "schema": {
+              "type": "integer"
+            },
+            "example": 10
+          },
+          {
+            "name": "X-Authorization-Token",
+            "in": "header",
+            "required": true,
+            "schema": {
+              "type": "string"
+            },
+            "example": "token"
+          },
+          {
+            "name": "filter[name]",
+            "in": "query",
+            "schema": {
+              "type": "object",
+              "properties": {
+                "name": {
+                  "type": "string"
+                }
+              }
+            },
+            "example": {
+              "name": "Example Table"
+            }
+          },
+          {
             "name": "filter[price]",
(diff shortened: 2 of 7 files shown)


lib/rspec/openapi/schema_merger.rb
```diff
@@ -25,22 +25,15 @@ def normalize_keys(spec)
...
       if base[key].is_a?(Hash) && value.is_a?(Hash)
-        deep_reverse_merge!(base[key], value)
-      elsif !base.key?(key)
-        base[key] = value
-      elsif base[key].is_a?(Array) && value.is_a?(Array)
-        # parameters need to be merged as if `name` and `in` were the Hash keys.
-        if key == 'parameters'
-          base[key] |= value
-          base[key].uniq! { |param| param.slice('name', 'in') }
+        if !base[key].key?("$ref")
```
exoego, line 32: This is a special treatment to avoid merge if the object is [Reference Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#reference-object).
The reference object can have only `$ref` field.
Invalid example:
```yaml
  schema:
     $ref: ".."
     description: "override"
```

spec/rails/doc/openapi.json
```diff
@@ -105,46 +105,6 @@
...
-              "properties": {
-                "name": {
-                  "type": "string"
-                }
-              }
-            },
-            "example": {
-              "name": "Example Table"
-            }
-          },
```
k0kubun, line 147: Could you explain the reason why `page`, `per`, `X-Authorization-Token`, and `filter[name]` need to be deleted?

exoego: Those were deleted because the below logic was deleted.

```
if key == 'parameters'
          base[key] |= value
          base[key].uniq! { |param| param.slice('name', 'in') }
```

Reverted the change.

exoego: Added a test case to ensure the base `parameters` are replaced with new spec. https://github.com/k0kubun/rspec-openapi/pull/55/commits/7bf6054ff36821201b842988562cb33cc015bd0e

lib/rspec/openapi/schema_merger.rb
k0kubun: Is this for the case that `$ref` is manually added to an OpenAPI schema file? If you manually maintain the schema of the response, why don't you skip updating OpenAPI from such test cases? Do you have a test case that calls multiple endpoints and want to update OpenAPI for only one of these?

lib/rspec/openapi/schema_merger.rb
```diff
@@ -25,22 +25,23 @@ def normalize_keys(spec)
...
+          deep_reverse_merge!(base[key], value)
+        end
       elsif base[key].is_a?(Array) && value.is_a?(Array)
         # parameters need to be merged as if `name` and `in` were the Hash keys.
         if key == 'parameters'
           base[key] |= value
           base[key].uniq! { |param| param.slice('name', 'in') }
+        else
+          base[key] = value
```
k0kubun, line 41: I wondered if just overwriting the value of `base` is fine, but given that what we have as an array in the current test in this repository are just `servers`, `tags`, `parameters`, and `example`, it's probably okay to rewrite it, ignoring the possibility of the need to partially update arrays inside it. We may need to recursively do something if we find other array fields that are not compatible with this idea, but I don't come up with such a thing at this moment.

lib/rspec/openapi/schema_merger.rb
```diff
@@ -25,22 +25,23 @@ def normalize_keys(spec)
...
         end
       else
-        # no-op
+        base[key] = value
```
k0kubun, line 44: Similar to https://github.com/k0kubun/rspec-openapi/pull/55/files#r877575816, as long as it updates only fields that rspec-openapi generates, it's probably fine, I guess.

k0kubun:
> This attempts to implement the behavior described in
> 
> > everything under the path/method should be erased while leaving keys that rspec-openapi cannot update as is,
and then changes generated by rspec-openapi should be added to it.

I'm okay with merging this patch, but note that the behavior of your implementation is not what you quoted. This PR simply doesn't implement "everything under the path/method should be erased". For example, when an endpoint currently returns `{ foo: 1, bar: 2 }` and it's changed to just `{ foo: 1 }`, rspec-openapi should assume the `bar` field is removed from the API and remove it from the existing file. But this patch doesn't do it. It only updates fields and never erases anything. So this PR actually doesn't close https://github.com/k0kubun/rspec-openapi/issues/48. 

Achieving it while still leaving legit manual changes isn't a simple task. It shouldn't be a simple recursive method, but it should be aware of which level of the schema it's currently at. That's why I called it "intelligent". 

But at this point, I'm not comfortable leaving https://github.com/k0kubun/rspec-openapi/issues/48 open since nobody would understand my intention of the current description. I don't afford to put so much effort on explaining it with concrete examples either. 

So, let's merge this and close https://github.com/k0kubun/rspec-openapi/issues/48. At least it brings us closer to what https://github.com/k0kubun/rspec-openapi/issues/48 says. Thanks.

PR #57, opened 2022-05-26
Much smarter merge: cleanup items that only exist in a previous result

exoego:
Addresses https://github.com/k0kubun/rspec-openapi/pull/55#issuecomment-1132259786

After this PR, some known items (below) will be removed if those exist only in a previous result.
So the removed URLs, parameters or such will be gone automatically.

1. `paths.*` URL paths
2. `paths.*.*` HTTP Methods
3. `paths.*.*.parameters`
4. `paths.*.*.requestBody.content.application/json.example.*`
5. `paths.*.*.requestBody.content.application/json.schema.properties.*`
6. `paths.*.*.responses.*.content.application/json.example.*`
7. `paths.*.*.responses.*.content.application/json.schema.properties.*`

Other items are preserved even if those exist only in a previous result.
So manually modified items can be retained.

I believe all patterns are covered by fixture-based test in `smart` directory.
Unit tests for internal implementations were removed not to bother refactoring in accordance with https://github.com/k0kubun/rspec-openapi/pull/57#discussion_r882335577

PR #58, opened 2022-05-26
Moved to rails-based testing

exoego:
This PR erases unit tests added in #55 and uses rails-rspec-based testing instead.
A small subset of existing OpenAPI file and test cases are added as test fixtures.

The below diff demonstrates how smart merge updates the existing fields.
```diff
❯ diff -u spec/rails/doc/smart/openapi.yaml spec/rails/doc/smart/expected.yaml
--- spec/rails/doc/smart/openapi.yaml   2022-05-27 06:51:14.000000000 +0900
+++ spec/rails/doc/smart/expected.yaml  2022-05-27 06:51:14.000000000 +0900
@@ -87,7 +87,7 @@
                     type: string
                     example: This field does not exist in rspec
               example:
-                message: This field should be updated by rspec
+                message: Unauthorized
     post:
       summary: create
       tags:
@@ -128,7 +128,7 @@
                 updated_at: '2020-07-17T00:00:00+00:00'
   "/tables/{id}":
     get:
-      summary: This filed should be updated by rspec
+      summary: show
       tags:
       - Table
       parameters:
```

The below diff indicates the current limitation of smart merge: it can not remove the non-existent field `no_such_field`.
I will add field removal in #57 later, based on this PR.

```diff
❯ diff -u spec/rails/doc/smart/openapi.yaml spec/rails/doc/openapi.yaml
--- spec/rails/doc/smart/openapi.yaml   2022-05-27 06:51:14.000000000 +0900
+++ spec/rails/doc/openapi.yaml 2022-05-27 06:51:14.000000000 +0900
(snip)
@@ -83,11 +106,8 @@
                 properties:
                   message:
                     type: string
-                  no_such_field:
-                    type: string
-                    example: This field does not exist in rspec
               example:
```

## Context
https://github.com/k0kubun/rspec-openapi/pull/57#discussion_r882335577

> Can you test it by only adding YAML-level tests at spec/rspec/rails_spec.rb? The only thing that must be maintained in this repository is what OpenAPI schema is generated at each rspec execution, not internal methods to implement it. If you unit-test internal implementations, it becomes harder to refactor the implementation.
>
> So I'm not comfortable having this kind of test when you could just write a test that fails only when needed. This test could fail when it doesn't need to, for example when we rename paths_to_all_fields to something else, which users don't care about. In fact, I actually don't appreciate tests under spec/rspec/schema_merger/ because the interface of RSpec::OpenAPI::SchemaMerger.merge! isn't part of the contact of this gem. Testing the same thing by reading a YAML file generated by running rspec would be more cost-effective because it would fail only when the actual contract is changed.


spec/rspec/rails_spec.rb
```diff
@@ -31,4 +31,26 @@
...
     end
   end
+
+  describe "smart merge" do
+    let(:openapi_path_for_smart_merge) do
+      File.expand_path('spec/rails/doc/smart/openapi.json', repo_root)
+    end
+
+    let(:expected_path_for_smart_merge) do
+      File.expand_path('spec/rails/doc/smart/expected.json', repo_root)
```
k0kubun, line 41: Can you just convert it to a normal YAML instead of a JSON?

exoego: Done in https://github.com/k0kubun/rspec-openapi/pull/58/commits/3f9398f7186e43f49e8b926ba886846f64f92ba8

k0kubun (approved):
Thanks :+1:

PR #59, opened 2022-05-27
Stabilize order of parameter objects and preserve objects with newer example

exoego:
This is a spin-out from #57
After this PR, `parameters` array preserves the occurrence with newer value in `example`

## Current implementation

It keeps the first occurrence of duplicated items with same `name` & `in`.
If user update an existing parameter in their rspec, a new parameter object with newer value in `example` field`is added to `parameters` array ... but is removed soon because it is not the first occurrence.
In other words, the occurrence with oldest example is kept forever.

## Problems

"Preserving first occurrence" will cause a problem for some cases.

### 1. User changed the value format.

Let's say user changed the format of `id` from numeric (`123`) to ulid (`01ARZ3NDEKTSV4RRFFQ69G5FAV`).

- Base spec, `[{ name: "id", in: "path", example: "123"  }]`
- After merge, `[{ name: "id", in: "path", example: "123"  }, { name: "id", in: "path", example:"01ARZ3NDEKTSV4RRFFQ69G5FAV"  }]`
- After deduplication, `[{ name: "id", in: "path", example: "123"  }]`

😭 The preserved `123` is wrong about format.

### 2. rspec-openapi support another field of parameter object

Let's say rspec-openapi added `description` field in parameter object somehow.

- Base spec, `[{ name: "id", in: "path", example: "123"  }]`
- After merge, `[{ name: "id", in: "path", example: "123"  }, { name: "id", in: "path", example:"123", description:"blah blah"  }]`
- After deduplication, `[{ name: "id", in: "path", example: "123"  }]`

😭 `description` is not added.

### 1. Keep a last occurrence 

Unfortunately, it brings unstable order. 
Let's say we invokes `/foo/123` and `/foo/456` in rspec.

- 1st run
  - Base spec, `[{ name: "id", in: "path", example: "123"  }]`
  - After merge, `[{ name: "id", in: "path", example: "123"  }, { name: "id", in: "path", example:"456"  }]`
  - After deduplication, `[{ name: "id", in: "path", example: "456"  }]`
- 2nd run: (just rerun)
  - Base spec, `[{ name: "id", in: "path", example: "456"  }]`
  - After merge, `[{ name: "id", in: "path", example: "456"  }, { name: "id", in: "path", example: "123"  } ]`
  - After cleanup, `[{ name: "id", in: "path", example: "123"  }]`

 😭 `123` and `456` is oscillating for each run. 

### 2. Introduces a timestamp marker to be used as sort key

I think it is better to introduce `__marker: Timestamp` field to `parameters`.
Then, merger sorts the parameters and preserves the duplicates with largest `__marker`, so the older occurrencenes will be removed.

This PR also sorts parameters by `in` and `name` before `__marker`, because `sort_by` is [not "stable sort"](https://ruby-doc.org/core-3.1.1/Enumerable.html#method-i-sort_by)

### DIFF
--- a/lib/rspec/openapi/schema_builder.rb
+++ b/lib/rspec/openapi/schema_builder.rb
@@ -48,7 +48,7 @@ class << RSpec::OpenAPI::SchemaBuilder = Object.new
   def build_parameters(record)
     parameters = []
 
-    marker_to_keep_last_duplicate = Time.now.utc.to_s
+    marker_to_keep_last_duplicate = Time.now.utc.round.to_s
 
     record.path_params.each do |key, value|
       parameters << {
--- a/lib/rspec/openapi/schema_builder.rb
+++ b/lib/rspec/openapi/schema_builder.rb
@@ -48,6 +48,8 @@ class << RSpec::OpenAPI::SchemaBuilder = Object.new
   def build_parameters(record)
     parameters = []
 
+    marker_to_keep_last_duplicate = Time.now.utc.to_s
+
     record.path_params.each do |key, value|
       parameters << {
         name: build_parameter_name(key, value),
@@ -55,6 +57,7 @@ class << RSpec::OpenAPI::SchemaBuilder = Object.new
         required: true,
         schema: build_property(try_cast(value)),
         example: (try_cast(value) if example_enabled?),
+        __marker: marker_to_keep_last_duplicate,
       }.compact
     end
 
@@ -64,6 +67,7 @@ class << RSpec::OpenAPI::SchemaBuilder = Object.new
         in: 'query',
         schema: build_property(try_cast(value)),
         example: (try_cast(value) if example_enabled?),
+        __marker: marker_to_keep_last_duplicate,
       }.compact
     end
 
@@ -74,6 +78,7 @@ class << RSpec::OpenAPI::SchemaBuilder = Object.new
         required: true,
         schema: build_property(try_cast(value)),
         example: (try_cast(value) if example_enabled?),
+        __marker: marker_to_keep_last_duplicate,
       }.compact
     end
 
--- a/lib/rspec/openapi/schema_builder.rb
+++ b/lib/rspec/openapi/schema_builder.rb
@@ -48,8 +48,6 @@ class << RSpec::OpenAPI::SchemaBuilder = Object.new
   def build_parameters(record)
     parameters = []
 
-    marker_to_keep_last_duplicate = Time.now.utc.round.to_s
-
     record.path_params.each do |key, value|
       parameters << {
         name: build_parameter_name(key, value),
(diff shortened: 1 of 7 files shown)


spec/rails/doc/smart/expected.yaml
```diff
@@ -48,12 +49,11 @@ paths:
...
-      - name: X-Authorization-Token
-        in: header
-        required: true
+      - name: page
+        in: query
         schema:
-          type: string
-        example: token
+          type: integer
+        example: 42
```
exoego, line 56: Note) `42` is newer value came from rspec

exoego:
Hmm, all tests passed on my end 🤔

spec/rails/doc/smart/expected.yaml
```diff
@@ -24,36 +24,36 @@ paths:
         in: query
         schema:
           type: integer
-        example: 1
+        example: 42
```
exoego, line 27: Note) `42` came from the updated test

k0kubun:
Sorry, I didn't read this when the CI was not passing and I didn't get a notification when it started to pass. I'll take a look.

lib/rspec/openapi/schema_builder.rb
```diff
@@ -48,13 +48,16 @@ def example_enabled?
   def build_parameters(record)
     parameters = []
 
+    marker_to_keep_last_duplicate = Time.now.utc.round.to_s
```
k0kubun, line 51: The PR is now concise enough to make me understand how and why it's used. Thank you.

I have some suggestions here though:

* Can we insert them inside `SchemaMerger` instead of `SchemaBuilder`? This seems to be the concern for `SchemaMerger` and it feels like the current implementation scatters the implementation details to unrelated places. This PR should touch `SchemaBuilder`.
* Could we achieve the same thing by just always prepending new schema elements to the array when merging them and calling `uniq!` after that? It should give the same result as sorting based on `__marker`.

exoego: > Can we insert them inside SchemaMerger instead of SchemaBuilder?

It makes sense. https://github.com/k0kubun/rspec-openapi/pull/59/commits/6326392058708f25889524086fb4d82f11f32d16
And it turned out that `__marker` is not needed. https://github.com/k0kubun/rspec-openapi/pull/59/commits/1128a0c0e60b8b12bd06c8c431f4b3d4ca4dd86e

> Could we achieve the same thing by just always prepending new schema elements to the array in SchemaMerger and calling uniq! after that?

Yes, if big diffs on the existing `rails/doc/openapi.(json|yml)` is acceptable.
https://github.com/k0kubun/rspec-openapi/pull/59/commits/2e88f7661a01181dc6994a71110e3c4cde7f379b

k0kubun (approved):
:+1:

PR #63, opened 2022-06-30
Accept empty array.

yasuzuki:
## Current Behaviour

When generating a schema with a response with an empty array, the item property is marked as nullable.

```yml
responses:
  '200':
    content:
      application/json:
        schema:
          type: array
          items:
            type: object
            properties:
              id:
                type: integer
              ...
            nullable: true
```

## Desired Behaviour

Should generate schema of typed as an array with the unknown property rather than nullable for the array itself.
Unknown property is expressed as `properties: {}` based on OAS 3.0.

reference: https://swagger.io/docs/specification/data-models/data-types/#array

When combined with other RSpec examples has a response with an array filled with any items, then generated schema should be merged correctly.

Thank you for this beautiful gem!

### DIFF
--- a/spec/requests/rails_spec.rb
+++ b/spec/requests/rails_spec.rb
@@ -42,6 +42,11 @@ RSpec.describe 'Tables', type: :request do
       end
     end
 
+    it 'can return an empty array' do
+      get '/tables', params: { filter: { "name" => "Never Match" } }, headers: { authorization: 'k0kubun' }
+      expect(response.status).to eq(200)
+    end
+
     it 'has a request spec which does not make any request' do
       expect(request).to eq(nil)
     end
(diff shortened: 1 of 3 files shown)


spec/requests/rails_spec.rb
```diff
@@ -42,6 +42,11 @@
       end
     end
 
+    it 'can return an empty array' do
+      get '/tables', params: { filter: { "name" => "Never Match" } }, headers: { authorization: 'k0kubun' }
+      expect(response.status).to eq(200)
+    end
```
exoego, line 48: Looks like this test case passes on the current `master` branch.
I am afraid that the changes are not covered.

exoego: I think it would be clearer if the behavior change is covered in the fixture files (JSON/YAML).
```
          items:
            type: object
            properties:
              id:
                type: integer
              ...
            nullable: true

===>


          items:
            type: object
            properties:
```

exoego:
@yasuzuki Thanks for proposing this change.
I've tweaked tests in #70 and merged. 
Closing as resolved.

yasuzuki:
@exoego Thank you for taking over our work!

PR #64, opened 2022-07-08
Use super_diff to make a tiny diff easily spottable

exoego:
While I was working on smart merge or other new feature proposal, it was a quite annoying to spot the actual diffs on two big hashes (OpenAPI definitions). 
This PR proposes to add https://github.com/mcmire/super_diff, as `:test` dependency,  which improves QoL on development.

## Before 
image

## After
image


k0kubun (approved):
Interesting, I didn't know of this. Using this looks good to me too :+1:

PR #74, opened 2022-08-25
Setup RuboCop

exoego:
To reduce efforts on code format and code review, let's introduce RuboCop.

This PR fixes some warning-level code offenses.
Other minor offenses will be addressed in subsequent PRs.


exoego:
I believe this eases us to maintain code quality.
We may configure RuboCop rules if we find them not so useful.

PR #78, opened 2022-11-02
Add request body merger

mocyuto:
I tried to add request body merger, because empty request body field is erased.

It works like as adding spec to this PR.


exoego (commented):
Use-case is not obvious to me.
Could you describe the issue you want to resolve and how this feature works for it.

mocyuto:
I thought POST request spec generator not merge but create new one.
But I tried to fix spec again, that fields were merged.

This PR is from my misunderstanding, so it would be close.

Thank you for reviewing!

PR #84, opened 2022-11-23
Support for security schemes/reqs

ksol:
👋🏼 This PR adds support for [specifying the security schemes used by the app](https://swagger.io/specification/#security-scheme-object) as well as setting them on [each operation](https://swagger.io/specification/#security-requirement-object)

It works as is, but I'm not sure what should be done with regards to the specs.

Let me know what I can do to make this PR mergeable!

### DIFF
--- a/spec/rails/doc/smart/expected.yaml
+++ b/spec/rails/doc/smart/expected.yaml
@@ -162,6 +162,12 @@ paths:
           type: integer
         example: 1
 components:
+  securitySchemes:
+    Scheme1:
+      description: Authentication scheme
+      type: http
+      scheme: bearer
+      bearerFormat: JWT
   schemas:
     Table:
       type: object
--- a/spec/requests/rails_smart_merge_spec.rb
+++ b/spec/requests/rails_smart_merge_spec.rb
@@ -15,6 +15,15 @@ RSpec::OpenAPI.comment = <<~COMMENT
   update this file automatically. You can also manually edit this file.
 COMMENT
 RSpec::OpenAPI.servers = [{ url: 'http://localhost:3000' }]
+RSpec::OpenAPI.security_schemes = {
+  'Scheme1' => {
+    description: 'Authentication scheme',
+    type: 'http',
+    scheme: 'bearer',
+    bearerFormat: 'JWT'
+  }
+}
+
 RSpec::OpenAPI.info = {
   description: 'My beautiful API',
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -146,6 +146,16 @@ RSpec::OpenAPI.response_headers = %w[X-Cursor]
 # Set `servers` - generate servers of a schema file
 RSpec::OpenAPI.servers = [{ url: 'http://localhost:3000' }]
 
+# Set `security_schemes` - generate security schemes
+RSpec::OpenAPI.security_schemes = {
```
exoego, line 150: Better to add test.

Please set `security_schemas` in https://github.com/k0kubun/rspec-openapi/blob/master/spec/requests/rails_smart_merge_spec.rb and update the corresponding openapi files accordingly.

lib/rspec/openapi/default_schema.rb
```diff
@@ -1,13 +1,22 @@
...
         version: RSpec::OpenAPI.application_version,
       },
       servers: RSpec::OpenAPI.servers,
       paths: {},
-    }.freeze
+    }
+
+    if RSpec::OpenAPI.security_schemes.present?
+      spec[:components] = {
+        securitySchemes: RSpec::OpenAPI.security_schemes
```
exoego, line 15: rspec-openapi does not touch "unknown" field added by users manually.
In other words, users can freely define securitySchemas in component block.
So I am inclined to reject this feature to reduce maintenance cost.

Could you elaborate why this feature is useful to users?

ksol: It's part of the OpenAPI spec, static-ish (unlike components.schemas which, it should neither change often, nor be dynamically defined).

Users that like their open api spec to document which authentication mechanism their endpoints use would benefit from this change by being able to define it in their app's code instead of having it to author the openapi file by hand

README.md
exoego, line 150: @ksol Could you add this in `Configuration` section in README?

lib/rspec/openapi/default_schema.rb
exoego: Understood.

README.md
exoego: Ah, sorry, THIS file is README 😓

exoego (approved):
LGTM
Thanks

Issue #88, opened 2023-01-27
Minitest support


oneiros:
First of all a big thank you to everyone involved in this fantastic project. I used to use `rspec_api_documentation`, which is sadly no longer maintained, and I believe one big reason is the many output formats it supports. Focusing on OpenAPI is the right way to go imho, so please keep up the good work!

I finally have a project I would really like to try this gem out on, but sadly it is using Rails' built-in test facilities, i.e. minitest, not rspec. I do not think I can really sell converting the existing test suite to rspec, so I almost gave up. But then I got this crazy (?) idea and started looking into what it would take to make this minitest-compatible.

As far as I can see, the only rspec-specific code is in `lib/rspec/openapi/hooks.rb`. With this is mind, I quickly hacked together a proof of concept, to see if porting this to minitest is feasible: https://github.com/oneiros/rspec-openapi/commit/54419044e963943d32886d8e25e5ebf70677fb64

It is really quick and dirty, but it _does_ work, kind of :grimacing: 

So now I wonder how I should proceed with this. I see the following possible paths:

1. If you are interested in providing minitest support from within this gem, I could polish my solution some more and prepare a pull request.
2. This gem could be split in two. One new gem providing the general facilities to generate OpenAPI files from data gathered during test runs. And this gem would then only provide the rspec specific stuff. Other people would be free to implement minitest-based solutions depending on the new gem.
3. I could fork this gem, rename it and rip out everyting rspec-specific.

What do you think?

exoego:
I would go with option 1. If a new version of rspec-openapi is published, minitest user also gain   benefits.
If it is separated as in option 2 or 3, other repo maintainers need to catch up.

Minitest is a 3rd most popular test framework in Rails ecosystem, so supporting it is highly welcomed from Rails community perspective.
https://rails-hosting.com/2022/#what-rails-testing-frameworks-are-you-using-to-write-tests![image](https://user-images.githubusercontent.com/127635/215267436-65261897-3d46-48c9-a5ef-0455b0aa71e3.jpeg)

Naming sounds a bit awkward if  :rspec-openapi supports minitest. We may give a better name to repo, but the name of gem will bb

exoego:
Closing as released in v0.8.0.

PR #90, opened 2023-02-22
Minitest support

oneiros:
It took me a while but I think I finally have something worth discussing. This PR builds upon my proof of concept mentioned in #88, streamlines the code a little bit and adds tests.

## Notable changes ##

* Handling the records created from the examples run is identical between both integrations. I opted for creating a new class, `ResultRecorder` to move that code into so it can be shared. This makes the actual integrations a lot slimmer, which I think is very nice.
* I need to hook into minitest in two very different ways. As such I cannot use a local outer variable as the rspec hooks do. So I created another attribute on the `RSpec::OpenAPI` module to collect the `path_records`. (A nice side-effect of the point mentioned above is that `error_records` has moved to the `ResultRecorder` class and is not needed as a global variable anymore.)

## Unsolved issues ##

* In contrast to my initial PoC I re-added the dependency on `rspec` in the `.gemspec` file. I had initially hoped that some kind of "soft dependency" would be possible, but there are a couple of issues with that. One is that `rspec-openapi` depends on and requires the `rspec` meta gem. `rspec-rails` on the other hand does not depend on that, so in many typical rails apps using `rspec` you will not find it in the gem bundle. I tried a couple of things but to no avail. If anyone has any ideas how one could remove the hard dependency, so minitest users will not have to install rspec, without any breaking changes for existing users, I would be thankful.
* I tried to mirror the existing specs for the rspec integration as much as possible. And initially I was very pleasantly surprised that I was able to generate almost the same openapi files using minitest. But I was never able to get the exact same output, because with the way `rspec-openapi` works the output heavily depends on the order in which the tests are executed. This is a problem with minitest, as it orders tests randomly during execution. The only other option is to order them alphabetically, which sadly does not match the order in the rspec request specs either. I had to patch minitest and reorder the tests to get the exact same output, which is not ideal imho. It makes the current tests _very_ brittle. OTOH I love that they prove that both integrations can produce the exact same output. So maybe it is OK.

## Missing features ##

My goal was to produce something usable that can serve as a basis for discussion so there are a couple of features I intentionally did not implement. Most notably:

* Per test-case metadata to override e.g. the description is not possible right now. Minitest has no concept of per test-case metadata, so this is rather hard to add. There are a couple of plugins for minitest that add similar functionality (e.g. tags), so it _could_ be done, but as stated I intentionally left this out for now.
* A custom `description_builder` will not work with minitest and I believe it never could because minitest has no concept of nested examples.
* ~~Using a `Proc` to construct the `path` will not currently work. This is actually not an intentional omission, but something I overlooked and only just noticed. I might still look into adding support for this.~~


oneiros:
> I tried to mirror the existing specs for the rspec integration as much as possible. And initially I was very pleasantly surprised that I was able to generate almost the same openapi files using minitest. But I was never able to get the exact same output, because with the way rspec-openapi works the output heavily depends on the order in which the tests are executed. This is a problem with minitest, as it orders tests randomly during execution. The only other option is to order them alphabetically, which sadly does not match the order in the rspec request specs either. I had to patch minitest and reorder the tests to get the exact same output, which is not ideal imho. It makes the current tests very brittle. OTOH I love that they prove that both integrations can produce the exact same output. So maybe it is OK.

Judging from the failed checks, I still do not have a way to guarantee a stable execution order of tests :frowning_face:

oneiros:
> Judging from the failed checks, I still do not have a way to guarantee a stable execution order of tests frowning_face

I made the ordering of tests explicit now, so it is 100% static. That should solve this issue.

Also, I added support for constructing the path dynamically using a `Proc`.

exoego (approved):
Sorry for late response.
LGTM

PR #110, opened 2023-04-28
Fix structural error - should NOT have fewer than 1 items

laerciosb:
First of all, the project is amazing and is serving me perfectly. I hope it continues to evolve for a long time.
Well, I found a problem when generating the openapi from the request of a `put` method. The required field is generated empty and this results in an error when reading the file as shown in the images below:

To fix the error, I simply removed the required field if it was empty.

### DIFF
--- a/lib/rspec/openapi/schema_cleaner.rb
+++ b/lib/rspec/openapi/schema_cleaner.rb
@@ -63,7 +63,7 @@ class << RSpec::OpenAPI::SchemaCleaner = Object.new
 
   def cleanup_hash!(base, spec, selector)
     RSpec::OpenAPI::HashHelper.matched_paths(base, selector).each do |paths|
-      base.dig(*paths).delete('required') if base.dig(*paths)['required']&.empty?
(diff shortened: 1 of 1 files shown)


exoego:
Thanks for the report and even a pull request!
The issue is reproducible with Swagger Editor on my side.

That place (inside `cleanup_hash!`) is a not place to be fixed.
`cleanup_hash!` used not only for `paths` field but also for many places, so adding that logic put unnecessary computation.

I think such empty array clean up added between `~~.cleanup!` and `~~update!` here:
https://github.com/exoego/rspec-openapi/blob/7fc9e4a981fe668717f9369191c48c9135144659/lib/rspec/openapi/result_recorder.rb#L27-L29
It would be great if you add test to verify new behavior.

I will look if you find it difficult.

exoego:
@laerciosb Thanks for your PR.
I've fixed the issue you report in https://github.com/exoego/rspec-openapi/pull/111 and published as `v0.8.1`.

Issue #56, opened 2022-05-20
Mounting engine to root "/" causes "No route matched for" in some scenarios


crnastena:
Any routes that are listed after mounting engine to root `mount Breaker::Engine => "/"` will generate `No route matched for` RuntimeError:

```
Failures:

  1) Api::V1::TestersController#index returns hello
     Failure/Error: raise "No route matched for #{request.request_method} #{request.path_info}"
     
     RuntimeError:
       No route matched for GET /testers
     # /Users/trikic/.rvm/gems/ruby-3.0.3/gems/rspec-openapi-0.5.0/lib/rspec/openapi/record_builder.rb:90:in `find_rails_route'
     # /Users/trikic/.rvm/gems/ruby-3.0.3/gems/rspec-openapi-0.5.0/lib/rspec/openapi/record_builder.rb:84:in `block in find_rails_route'
     # /Users/trikic/.rvm/gems/ruby-3.0.3/gems/rspec-openapi-0.5.0/lib/rspec/openapi/record_builder.rb:81:in `find_rails_route'
     # /Users/trikic/.rvm/gems/ruby-3.0.3/gems/rspec-openapi-0.5.0/lib/rspec/openapi/record_builder.rb:21:in `build'
     # /Users/trikic/.rvm/gems/ruby-3.0.3/gems/rspec-openapi-0.5.0/lib/rspec/openapi/hooks.rb:14:in `block in <top (required)>'

Finished in 0.26384 seconds (files took 4.66 seconds to load)
1 example, 1 failure
```

Here is a sample `routes.rb`:
```ruby
  namespace :api do
    namespace :v1 do

      # gives errors
      # mount Breaker::Engine => "/", as: :breaker
      # mount Breaker::Engine => "/"

      # works
      # mount Breaker::Engine => "/breaker"

      resources :testers, only: [:index]

      # works
      mount Breaker::Engine => "/"
    end
  end

```

Full project that reproduces the error can be found at [crnastena/openapi-test](https://github.com/crnastena/openapi-test).

I guess this more of am enquiry as I am not sure if it is a problem with `rspec-openapi` or not. What do you think?

k0kubun:
Thanks for preparing a repository. That is really helpful.

[...]

PR #113, opened 2023-05-05
Fix engine path resolution

pcoliveira:
Closes issue #56 
BREAKING CHANGE: The returned path for mounted engines now includes the path prefix.

An error was raised when testing a route that is mounted after an engine that shares the same route prefix.
The path returned for a mounted engine didn't include the route prefix (only the engine sub path was returned).

On the engine test case, the expected path was changed from `/eng_route` to `/my_engine/eng_route`, to match the mounted path.


xcskier56:
@pcoliveira is there anything I can do to help this PR along? This has been nagging me for a long while and finally has become more of an issue

lib/rspec/openapi/record_builder.rb
```diff
@@ -99,21 +107,18 @@ def rack_test?(context)
...
-        return route
+        if route.app.engine?
+          route, path = find_rails_route(request, app: route.app.app, path_prefix: path)
+          next if route.nil?
+        end
+        return [route, path_prefix + path]
       end
     end
-    raise "No route matched for #{request.request_method} #{request.path_info}"
+    nil
```
exoego, line 121: Q) Why this error message is removed?

argent-codes:
My organization recently ran into the issue fixed by this PR, so I would love to see this released ❤️

exoego:
@pcoliveira Hi. Sorry for late response. 
Could you check your PR?
I left a question.

lib/rspec/openapi/record_builder.rb
pcoliveira: When two engines are mounted on the same root path, and the first doesn't resolve, the resolution needs to continue to the next engine. The raise was moved to the caller, when all engines are resolved

pcoliveira:
> @pcoliveira Hi. Sorry for late response. Could you check your PR? I left a question.

sorry, I missed your question.

pcoliveira:
@exoego How may I help? is there anything I need to do to move the PR forward?

exoego (approved):
LGTM
This is amazing.
Sorry for the late response.

PR #114, opened 2023-05-12
Add required request params to metadata

Alexey2257:
This PR adds ability to specify required request params in metadarta


spec/rails/doc/smart/expected.yaml
```diff
@@ -28,16 +28,19 @@ paths:
...
         example: 42
       - name: per
         in: query
+        required: false
         schema:
           type: integer
         example: 10
       - name: show_columns
         in: query
+        required: false
```
exoego, line 43: @Alexey2257 Could you update one of the test case to include `required_request_params: [ 'show_columns']` so we ensure `required: true` is generated?

exoego:
Thanks for your contribution!
With this change, `required: false` will be added to existing query parameters unless users specify `required_request_params`.
That new behavior might be a breaking change.
But I assume that new behavior is preferred since query parameters are optional in most cases, IMHO.

I requested a small change.

spec/rails/doc/smart/expected.yaml
Alexey2257: Sure, done

Alexey2257:
> Thanks for your contribution! With this change, `required: false` will be added to existing query parameters unless users specify `required_request_params`. That new behavior might be a breaking change. But I assume that new behavior is preferred since query parameters are optional in most cases, IMHO.
> 
> I requested a small change.

Thanks for your review! Adding "required: false" everywhere shouldn't be breaking change because OpenApi specification suggests that this parameter is default to "false" when omitted.  But if you wish to preserve previous behavior we can change code to something like this so required key wouldn't be present when it is not true

``` Ruby
parameters << {
  name: build_parameter_name(key, value),
  in: 'query',
  schema: build_property(try_cast(value)),
  example: (try_cast(value) if example_enabled?),
}.tap do |hash|
  hash[:required] = true if record.required_request_params.include?(key)
end.compact
```

exoego (approved):
Sorry for late response.
LGTM 👍

PR #117, opened 2023-05-21
Fix multiple uploaded files examples

t0yohei:
Adjust the multiple example parameters.

I get this linting error with an endpoint that receives an uploaded file: unknown tag !<!ruby/object:File>

The spec is structured in this way:

```
  describe "POST /api/v1/my_pages" do
    subject do
      post api_v1_my_pages_path(my_page), params: params, headers: default_headers
    end

    let!(:user) { create(:user) }
    let!(:my_page) { create(:my_page) }
    let(:params) { { example: { title: 'test', images: [fixture_file_upload('/images/Logo.png')] } } }

      it 'success' do
        sign_in user
        expect(response).to have_http_status(:created)
      end
```

I have added the fix for this issue by speculating that the cause lies in the schema_builder.rb file's failure to handle cases where image is sent as an array.

### DIFF
--- a/lib/rspec/openapi/schema_builder.rb
+++ b/lib/rspec/openapi/schema_builder.rb
@@ -195,6 +195,11 @@ class << RSpec::OpenAPI::SchemaBuilder = Object.new
         value[key] = v.original_filename
       elsif v.is_a?(Hash)
         adjust_params(v)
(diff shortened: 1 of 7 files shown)


exoego (approved):
LGTM
Thanks

PR #126, opened 2023-07-06
Fix arrayed uploaded file examples

champierre:
I created this Pull Request because I encountered a problem similar to https://github.com/exoego/rspec-openapi/pull/46. Specifically, when I included an array of hashes in the POST request parameters, where the hash values were image files, I couldn't load the generated yaml file.


exoego:
LGTM
Thanks

PR #127, opened 2023-07-26
Don't dump records into temporary file

andyundso:
I browsed the source code the other day and found this statement. As it is referenced nowhere else in the code, I assume it was used for debugging purposes when implementing the Minitest support and is therefore no longer necessary.


exoego:
Thanks.
It was added in https://github.com/exoego/rspec-openapi/pull/90/files#diff-61df5b43e87c7076a0f7eeda68626183021a685cd305969c9f49c940fa6c7c7fR16 and no longer needed.

PR #128, opened 2023-07-26
Prepend `openapi!` in any case

andyundso:
If you run tests with Minitest at the moment, but without the environment variable `OPENAPI`, you are greeted with an error message.

```
undefined method `openapi!' for Api::V0::SomeControllerTest:Class (NoMethodError)`
```

This is because the current code adds the `openapi!` method only if `OPENAPI` environment variable is set.

This commit modifies the loading logic, always requiring the `minitest` hooks class to be loaded if `Minitest` constant is defined, to avoid issues with RSpec. Then, the `openapi?` and `openapi!` methods are always prepended to `Minitest::Test`, while the modified `run` method and the `after_run` hook are only activated when `OPENAPI` environment variable is set.

### DIFF
--- a/lib/rspec/openapi.rb
+++ b/lib/rspec/openapi.rb
@@ -10,10 +10,8 @@ require 'rspec/openapi/schema_file'
 require 'rspec/openapi/schema_merger'
 require 'rspec/openapi/schema_cleaner'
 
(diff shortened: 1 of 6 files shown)


exoego:
Thanks. Makes sense to me.

PR #131, opened 2023-08-17
added operationId to metadata of request specs

jamerine:
Added a change to allow users to add operationId to the metadata of request specs. There have been requests to make the  OpenAPI spec require operationId for many reasons. Here is an example: https://github.com/OAI/OpenAPI-Specification/issues/1907. 

If a user adds `operation-id` to the metadata of a request spec, it will be added to the path section of the schema.  

First time contributor and I struggled to figure out where to add a test for this feature. Let me know if there is a place where you would like to see a spec for this.

### DIFF
--- a/lib/rspec/openapi/record.rb
+++ b/lib/rspec/openapi/record.rb
@@ -11,6 +11,7 @@ RSpec::OpenAPI::Record = Struct.new(
   :request_headers,       # @param [Array]  - [["header_key1", "header_value1"], ["header_key2", "header_value2"]]
   :summary,               # @param [String]  - "v1/statuses #show"
   :tags,                  # @param [Array]   - ["Status"]
+  :operation_id,          # @param [String]   - "request-1234"
   :description,           # @param [String]  - "returns a status"
   :security,              # @param [Array]  - [{securityScheme1: []}]
   :status,                # @param [Integer] - 200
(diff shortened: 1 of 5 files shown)


jamerine:
@exoego - What are your thoughts on this feature to enable `operationId` to be passed in as metadata options on a spec?

exoego:
Supporting `operation-id` itself sounds legit to me.
It is unclear to me that `path_server_records` or other props are added.

jamerine:
> Supporting `operation-id` itself sounds legit to me. It is unclear to me that `path_server_records` or other props are added.

@exoego - Sorry for the confusion here, I was testing out another proposed feature to the gem and committed my change to this PR. I am going to revert my last commit here so that this PR is strictly for the `operation-id` feature. Then I'll submit another PR for the server record changes.

jamerine:
@exoego - Ok, this PR for adding `operation-id` is good to review again.

exoego (commented):
@jamerine Could you update any single test to use `:operation_id` so we can ensure  OpenAPI file generated by tests have `operationId`.

jamerine:
@exoego - We added `:operation_id` to  a test and updated the expected.yaml to have the operationId as well. Let me know if there are any other changes you would like us to make.

jamerine:
@exoego - If we can get this `:operation_id` PR merged in, I will open up another PR for another proposed new feature.

exoego (approved):

exoego:
Sorry for the late response.
Thanks.

PR #148, opened 2023-11-30
Add ability to define custom summary and tags builders

mercedesb:
Love this project! We use it at my company for automated documentation (using `widdershins` to take the generated OpenAPI spec and create nice Slate docs). So thank you!

We'd love to have some control over how we decide what the summary and tags are for a given example. 

For our use case, we want to be able to define tags on the parent example and have that applied to all the child examples rather than having to define them on every example.

I followed the existing pattern in the repo for custom descriptions so that any custom scenario could be supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,10 +170,12 @@ EOS
 RSpec::OpenAPI.description_builder = -> (example) { example.description }
 
 # Generate a custom summary, given an RSpec example
-RSpec::OpenAPI.summary_builder = -> (example) { example.metadata[:summary] }
+# This example uses the summary from the example_group.
+RSpec::OpenAPI.summary_builder = ->(example) { example.metadata.dig(:example_group, :openapi, :summary) }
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -167,6 +167,12 @@ EOS
 # Generate a custom description, given an RSpec example
 RSpec::OpenAPI.description_builder = -> (example) { example.description }
 
+# Generate a custom summary, given an RSpec example
+RSpec::OpenAPI.summary_builder = -> (example) { example.metadata[:summary] }
```
exoego, line 171: Could you give a short example here, to give an idea of customization?

exoego (commented):
@mercedesb 
Thanks.
The feature makes sense to me.
I think a short example of customization is helpful for other users.

exoego (approved):
LGTM
Thanks

PR #150, opened 2023-11-30
Add ability to configure which path params to ignore

mercedesb:
It would be nice to be able to configure which path params to ignore. For example, sometimes it would be nice to remove path and subdomain.

By default, the configuration hasn't changed but this adds the ability to change it if you need more flexibility.


exoego (approved):
LGTM
Thanks

Issue #161, opened 2023-12-13
Using `RSpec::OpenAPI.path = ->` but for different audiences


ipepe:
Contextual snippet:
```ruby
RSpec::OpenAPI.path = -> (example) {
  case example.file_path
  when %r[spec/requests/api/v1/] then 'doc/openapi/v1.yaml'
  when %r[spec/requests/api/v2/] then 'doc/openapi/v2.yaml'
  else 'doc/openapi.yaml'
  end
}
```

I really like the idea behind generating multiple schemas, but the current configuration causes a caveat that makes this feature challenging to use. (I didn't test this in practice but that is what I assume happens based on documentation)

It's because these different schema files, which might be targeted at different audiences are sharing common values for other configuration options like: 

```ruby
RSpec::OpenAPI.title = 'OpenAPI Documentation'
RSpec::OpenAPI.enable_example = false
RSpec::OpenAPI.application_version = '1.0.0'
RSpec::OpenAPI.info = {
  description: 'My beautiful API',
  license: {
    'name': 'Apache 2.0',
    'url': 'https://www.apache.org/licenses/LICENSE-2.0.html'
  }
}
RSpec::OpenAPI.request_headers = %w[X-Authorization-Token]
RSpec::OpenAPI.response_headers = %w[X-Cursor]
RSpec::OpenAPI.servers = [{ url: 'http://localhost:3000' }]
RSpec::OpenAPI.security_schemes = {
  'MyToken' => {
    description: 'Authenticate API requests via a JWT',
    type: 'http',
    scheme: 'bearer',
    bearerFormat: 'JWT',
  },
}
RSpec::OpenAPI.comment = 'comment'
RSpec::OpenAPI.description_builder = -> (example) { example.description }
RSpec::OpenAPI.summary_builder = ->(example) { example.metadata.dig(:example_group, :openapi, :summary) }
RSpec::OpenAPI.tags_builder = -> (example) { example.metadata.dig(:example_group, :parent_example_group, :openapi, :tags) }
RSpec::OpenAPI.example_types = %i[request]
RSpec::OpenAPI.ignored_path_params = %i[controller action format]
```

I don't have a suggestion on how this should be changed. But currently, I cannot generate schema for our frontend API team that is internal and second one for external and public API schema for our consumers.

exoego:
Let me confirm.
Do you want to customize `title`, `info`, `comment` or such based on pathes, for different audiences?
Something like #162?

PR #162, opened 2023-12-13
Allow path-based config overrides

cyrusg:
Closes https://github.com/exoego/rspec-openapi/issues/161
----

Complex API's often use different server paths and have specific title and other information for different services. The purpose of this PR is to allow for such configurations to be overridden within each path defined in `RSpec::OpenAPI.path`. The PR also allows for the naming of the config file (defaulting to `rspec_openapi.rb`).

While the current initializer can define all the paths and common settings like the request and response headers, security schemes, and others, the path-based configuration files can be used to define server and other path/service-specific information.

### DIFF
--- a/lib/rspec/openapi.rb
+++ b/lib/rspec/openapi.rb
@@ -26,6 +26,7 @@ module RSpec::OpenAPI
   @example_types = %i[request]
   @response_headers = []
   @path_records = Hash.new { |h, k| h[k] = [] }
+  @config_filename = 'rspec_openapi.rb'
 
   class << self
     attr_accessor :path,
@@ -39,6 +40,7 @@ module RSpec::OpenAPI
                   :security_schemes,
                   :example_types,
                   :response_headers,
-                  :path_records
+                  :path_records,
+                  :config_filename
   end
 end
--- a/lib/rspec/openapi.rb
+++ b/lib/rspec/openapi.rb
@@ -26,6 +26,8 @@ module RSpec::OpenAPI
   @example_types = %i[request]
   @response_headers = []
   @path_records = Hash.new { |h, k| h[k] = [] }
+
+  # This is the configuraion override file name we look for within each path.
   @config_filename = 'rspec_openapi.rb'
 
   class << self
@@ -40,7 +42,8 @@ module RSpec::OpenAPI
                   :security_schemes,
                   :example_types,
                   :response_headers,
-                  :path_records,
-                  :config_filename
+                  :path_records
+
+    attr_reader   :config_filename
   end
 end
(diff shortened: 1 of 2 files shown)


exoego:
> Complex API's often use different server paths and have specific title and other information for different services. 

Could you let me know?
Do you want to use multiple config filed so you can create multiple OpenAPI files based on paths, something like:

```ruby
# open_api_v1.rb
RSpec::OpenAPI.title = 'API v1 (legacy, unmaintained)'
RSpec::OpenAPI.path = 'doc/schema_v1.yaml'


# open_api_v2.rb
RSpec::OpenAPI.title = 'API v2'
RSpec::OpenAPI.path = 'doc/schema_v2.yaml'
```

If so, `RSpec::OpenAPI.path` allows splitting the files based on paths.

```ruby
RSpec::OpenAPI.path = -> (example) {
  case example.file_path
  when %r[spec/requests/api/v1/] then 'doc/openapi/v1.yaml'
  when %r[spec/requests/api/v2/] then 'doc/openapi/v2.yaml'
  else 'doc/openapi.yaml'
  end
}
```

cyrusg:
> Could you let me know? Do you want to use multiple config filed so you can create multiple OpenAPI files based on paths, something like:
> 
> ```ruby
> # open_api_v1.rb
> RSpec::OpenAPI.title = 'API v1 (legacy, unmaintained)'
> RSpec::OpenAPI.path = 'doc/schema_v1.yaml'
> 
> 
> # open_api_v2.rb
> RSpec::OpenAPI.title = 'API v2'
> RSpec::OpenAPI.path = 'doc/schema_v2.yaml'
> ```
> 
> If so, `RSpec::OpenAPI.path` allows splitting the files based on paths.
> 
> ```ruby
> RSpec::OpenAPI.path = -> (example) {
>   case example.file_path
>   when %r[spec/requests/api/v1/] then 'doc/openapi/v1.yaml'
>   when %r[spec/requests/api/v2/] then 'doc/openapi/v2.yaml'
>   else 'doc/openapi.yaml'
>   end
> }
> ```

Basically yes, that's what we're looking for. But my understanding is that path splitting just tells the gem where to dump the schemas for each API. We need to be able to change the title, server locations, etc. I think this is also what ipepe is asking for in #161.

A little more context: We are a benefits company and have several API's such as enrollments, mobile, etc. They all live in the same base repo, so we somehow have to be able to generate documentation with the right titles, etc. for each of them. 

Right now, we have a config file in our `config/initializers` folder that has most of the common setting across our API's. But each API has a different title, description, and server path(s). And that's the impetus for this PR.

exoego:
@cyrusg Thanks.
I've added some commits in https://github.com/exoego/rspec-openapi/pull/165 and merged your PR.

PR #163, opened 2023-12-14
Sort HTTP methods, response status codes, and contents lexicographically

### LINKED ISSUE #159: Regenerating (remove file and run rspec) the schema without changing the code, generates different file.
Regenerating (remove file and run rspec) the schema without changing the code, generates different file. I assume this is because RSpec is executing tests in different order each time and this causes schema to be generated differently. Would be nice to have some counter-measures for this like:
 - http status code keys are ordered in ascending manner
   - this should prevent code 400 and 500 to be in different order each time based on rspec randomization of order of tests
 - any schema keys are ordered alphabetically
   - this should prevent situations where payload is being extended by one or other test

exoego:
Closes #159 

# Changes

This is a generalization of #155 

The below keys are sorted lexicographically.

- API Paths. E.g.) `/users/`, `/users/{id}`. This is already implemented in #155, but refactored.
- HTTP methods. E.g.) `get`, `delete`, `post`, `put`
- Response status codes. E.g.) `200`, `301`, `400`, `500`
- Response content-types: E.g.) `application/json`, `text/plain`

Keys in other places remain as-is.

# Alternative considered
I initially implemented sorting every `Hash`, but it is not good.
Keys of some places should remain as-is.

E.g.) Some top-level keys. It makes a bit surprising if`openapi:` keys are in the very bottom of the file. 

```yaml
openapi: 3.0.3
info:
  title: OpenAPI Documentation
  version: 1.0.0
servers: []
paths:
```

E.g.) The properties of schema. I assume users defined properties as they feel it natural.
```yaml
  schemas:
    Table:
      type: object
      properties:
        id:
          type: integer
        # This field should exists in expected
        # name:
        #  type: string
        this_field_should_not_exist_in_expected:
          type: string
        description:
          type: string
        database:
          "$ref": "#/components/schemas/Database"
        columns:
          type: array
          items:
            "$ref": "#/components/schemas/Column"
        null_sample:
          nullable: true
        storage_size:
          type: number
          format: float
        created_at:
          type: string
        updated_at:
          type: string
```

### DIFF
--- a/lib/rspec/openapi.rb
+++ b/lib/rspec/openapi.rb
@@ -9,6 +9,7 @@ require 'rspec/openapi/schema_builder'
 require 'rspec/openapi/schema_file'
 require 'rspec/openapi/schema_merger'
 require 'rspec/openapi/schema_cleaner'
+require 'rspec/openapi/schema_sorter'
 
 require 'rspec/openapi/minitest_hooks' if Object.const_defined?('Minitest')
 require 'rspec/openapi/rspec_hooks' if ENV['OPENAPI'] && Object.const_defined?('RSpec')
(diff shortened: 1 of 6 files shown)


ipepe:
@exoego I'm using this version and I noticed that it doesn't sort properly. Seems like files that were already generated with the wrong sorting are not being sorted after the update to this version?

PR #167, opened 2024-01-10
Correctly normalize rails paths with optional bound params

mercedesb:
Rails supports [optional bound parameters](https://guides.rubyonrails.org/routing.html#bound-parameters).

When using optional params, the generated paths by this gem look something like: `/optional_params/(/{id)}` instead of the expected `/optional_params/(/{id})`

If you use tools like [widdershins](https://www.npmjs.com/package/widdershins) to generate docs from your OpenAPI specification, this can break the parsing.

This PR updates the regex to not include the closing parentheses in the capture group during path normalization.

### DIFF
--- a/spec/rails/doc/openapi.json
+++ b/spec/rails/doc/openapi.json
@@ -341,6 +341,48 @@
         }
       }
     },
+    "/optional_params(/{id})": {
+      "get": {
+        "summary": "display",
+        "tags": [
+          "OptionalParam"
+        ],
+        "responses": {
+          "200": {
+            "description": "returns a payload",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "properties": {
+                    "id": {
+                      "type": "string"
+                    }
+                  },
+                  "required": [
+                    "id"
+                  ]
+                },
+                "example": {
+                  "id": "17"
+                }
+              }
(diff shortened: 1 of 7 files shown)


spec/rails/doc/openapi.json
```diff
@@ -341,6 +341,48 @@
         }
       }
     },
+    "/optional_params(/{id})": {
```
exoego, line 344: @mercedesb Could you check OpenAPI documentation to verify `"/optional_params(/{id})"` is valid format of optional path parameter?
I've looked https://swagger.io/docs/specification/paths-and-operations/
 but could not find

mercedesb: Ahhh good call, it does not. [Here's an open issue on the OpenAPI Specification repo where people are asking for it](https://github.com/OAI/OpenAPI-Specification/issues/93). I'll close this. Thanks for asking this question! I didn't think to check this.

mercedesb:
[The official OpenAPI Spec does not support optional path parameters](https://github.com/OAI/OpenAPI-Specification/issues/93)

mercedesb:
[The official OpenAPI Spec does not support optional path parameters](https://github.com/OAI/OpenAPI-Specification/issues/93)

Issue #173, opened 2024-02-06
Manually edited oneOf in response schema is overridden


simba909:
I'm evaluating this gem for a project I'm working on and it looks _much_ nicer than rswag. Well done! 👏🏻 

Sorry for the long message that follows.

Now, my Rails app has a model that itself has a property that can be an instance of several different sub-models. The JSON for this can look something like:
```json
{
  "name": "some_name",
  "data": {
    "email": "someone@example.com"
  }
}
```

or:
```json
{
  "name": "some_name",
  "data": {
    "selfie_url": "https://somewhere.com/public/image.jpg"
  }
}
```

I have request specs that return both variants of this model. By default rspec-openapi generates an OpenAPI spec that looks something like this:
```yaml
components:
  schemas:
    MyModel:
      type: object
      properties:
        name:
          type: string
        data:
          type: object
          properties:
            email:
              nullable: true
            selfie_url:
              nullable: true
```

Which is fine but not quite what I want. I'd like to use the `oneOf` keyword to differentiate the response types. So I edit my OpenAPI spec to look like this:
```yaml
components:
  schemas:
    MyModel:
      type: object
      properties:
        name:
          type: string
        data:
          oneOf:
            - type: object
              properties:
                email:
                  type: string
                  format: email
            - type: object
               properties:
                 selfie_url:
                   type: string
                   format: uri
```

But when I rerun the spec generation, rspec-openapi alters the spec to this:
```yaml
components:
  schemas:
    MyModel:
      type: object
      properties:
        name:
          type: string
        data:
          oneOf:
            - type: object
              properties:
                email:
                  type: string
                  format: email
            - type: object
               properties:
                 selfie_url:
                   type: string
                   format: uri
            type: object
            properties:
              email:
                nullable: true
              selfie_url:
                nullable: true
```

Notice that it keeps my `oneOf:` but adds back the sub-model with all properties being `nullable`. Is this a known limitation or something I'm doing wrong?

hss-mateus:
Same issue here. Reading through the source code it seems that this case isn't handled.
A possible implementation is to find the most similar schema, merge it, and if no schema is sufficiently similar, a new option is added to `oneOf`.

exoego:
```yaml
        data:
          oneOf:
            - type: object
              properties:
                email:
                  type: string
                  format: email
            - type: object
               properties:
                 selfie_url:
                   type: string
                   format: uri
            type: object
            properties:
              email:
                nullable: true
              selfie_url:
[...]

PR #174, opened 2024-02-08
Merge manually added `oneOf` schemas

hss-mateus:
As reported in #173, manually added `oneOf` attributes are ignored and the generated schema is placed alongside.

This PR implements an algorithm that compare the generated schema with every `oneOf` option, merges with the most similar one, or adds a new option if none are sufficiently similar.

I placed an abitrary value of 0.5 that represents the minimum similarity score required.

I have to update the `$ref` handling code too, since it was discarding paths inside arrays.

### DIFF
--- a/lib/rspec/openapi/components_updater.rb
+++ b/lib/rspec/openapi/components_updater.rb
@@ -74,6 +74,7 @@ class << RSpec::OpenAPI::ComponentsUpdater = Object.new
     nested_refs = [
       *RSpec::OpenAPI::HashHelper.matched_paths_deeply_nested(base, 'components.schemas', 'properties.*.$ref'),
       *RSpec::OpenAPI::HashHelper.matched_paths_deeply_nested(base, 'components.schemas', 'properties.*.items.$ref'),
+      *RSpec::OpenAPI::HashHelper.matched_paths_deeply_nested(base, 'components.schemas', 'oneOf.*.$ref'),
     ]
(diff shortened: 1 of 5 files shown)


simba909:
I've tested this with my setup from #173 and it works beautifully, well done @hss-mateus! 👏🏻 

@exoego Have you had a chance to look at this yet?

exoego:
> This PR implements an algorithm that compare the generated schema with every oneOf option, merges with the most similar one, or adds a new option if none are sufficiently similar.
>
> I placed an abitrary value of 0.5 that represents the minimum similarity score required.

That algorithm sounds smart.
Please add a test case for this behavior in `requests/rails_smart_merge_spec.rb`.
A new option of `oneOf`, which is not found in `rails/doc/smart/openapi.yaml`, should be added to `rails/doc/smart/expected.yaml`.

hss-mateus:
> Please add a test case for this behavior in `requests/rails_smart_merge_spec.rb`. A new option of `oneOf`, which is not found in `rails/doc/smart/openapi.yaml`, should be added to `rails/doc/smart/expected.yaml`.

If I understand it correctly, [this](https://github.com/exoego/rspec-openapi/pull/174/files#diff-875f9a36fc09b81250e980b5767df84ae8b967a83ab5919c65914b21a02b2ec3R148-R156) part of the generated schema is already not found in `openapi.yml`, but is being added to `expected.yml`

exoego:
Ah, I overlooked that.
Thanks.

PR #178, opened 2024-03-01
Additional properties override

nicvw:
A possible implementation to address [Issue 132](https://github.com/exoego/rspec-openapi/issues/132).  This will not add a `properties` or `required` field if `additionalProperties` exists on the base schema.  It will still update those fields if they are already present on the base, I feel that is a decent compromise to resolve that issue and not interfere with schemas where all three fields are intentionally present.


exoego:
Thanks!

exoego:
@nicvw Thanks for your contribution. This was released in v0.13.0.

PR #197, opened 2024-03-28
normali

exoego:
During #196, I found string and symbols are mixed up.

String keys come from the base file (`YAML.safe_load(File.read(@path))`) and Symbol keys come from the generate specs.
Since String != Symbol, merge is buggy.


exoego:
Superseded by https://github.com/exoego/rspec-openapi/pull/198

PR #202, opened 2024-03-28
Hanami support

AlexeyMatskevich:
As part of this MR, I implemented initial support for Hanami version 2. 

Basically I provided a Hanami application, which is a re-implementation of the Rails application, with the only difference that instead of rails engine, I provided [hanami slice](https://guides.hanamirb.org/v2.1/app/slices/).

I wrote tests for hanami similarly to the tests for rails.

I have not implemented the tests on minitest as I don't know it.

If you would be interested we could finalise this MR, infuse, and after that within the new MRs I would sea implement the rest of the features.

Of the things that aren't working right now, I can see
1. it doesn't work to define id as a parameter, respectively describe it in openapi and merge all similar queries such as `/table/1`, `/table/2`.
2. As a consequence of the first point, different answers at different id, are in different descriptions, a grouping problem.
3. No tags
4. Summury is generated in a different way from rails

Key changes:
1. Moved all test apps to spec/apps
2. Removed the use of index_by, as it is part of activesupport


exoego (commented):
Awesome 🥇 
The change itself LGTM.
Could you fix conflicts?

lib/rspec/openapi/record_builder.rb
```diff
@@ -55,7 +55,10 @@ def safe_parse_body(response, media_type)
   def extract_headers(request, response)
     request_headers = RSpec::OpenAPI.request_headers.each_with_object([]) do |header, headers_arr|
       header_key = header.gsub('-', '_').upcase.to_sym
-      header_value = request.get_header(['HTTP', header_key].join('_')) || request.get_header(header_key)
+
+      header_value = request.get_header(['HTTP', header_key].join('_')) ||
+                     request.get_header(header_key) ||
+                     request.get_header(header_key.to_s)
```
AlexeyMatskevich, line 61: I found one problem with the new edits - https://github.com/exoego/rspec-openapi/pull/198/files#diff-a79ceb062f9ec26b6203f0b7541bd0a9b1cc2ebf3370cac3da9039f4f11dc1a7R56 - Rack test seems to send strings instead of characters in headers, so because of this edit, the headers are no longer defined.

exoego:
@AlexeyMatskevich 
The job for Ruby 2.7 is failing at bundle install step: https://github.com/exoego/rspec-openapi/actions/runs/8498580117/job/23302368385?pr=202

It appears that Hanami 2.0 dropped Ruby 2.7 support https://github.com/hanami/hanami/blob/main/CHANGELOG.md#changed-14

Please install Hanami v1.x when Ruby 2.7, or exclude Hanami test when Ruby 2.7.
Either is fine.
Ruby 2.7 reached EOL, but some users may still use Hanami 1.x and Ruby 2.7.

AlexeyMatskevich:
Hanami version 1 and version 2 are essentially two different frameworks, they are not compatible, so I would prefer to just exclude the tests for version 2.7, but I don't see how to implement that yet.

spec/rspec/hanami_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+require 'spec_helper'
+require 'yaml'
+require 'json'
+require 'pry'
+
+RSpec.describe 'hanami request spec' do
+  include SpecHelper
+
+  describe 'yaml output' do
```
exoego, line 13:
```suggestion
  include SpecHelper

  # Hanami 2.0 support Ruby 3.0+
  skip if RUBY_VERSION < '3.0.0'

  describe 'yaml output' do
```

AlexeyMatskevich: I think I've succeeded

exoego:
Thanks!

PR #205, opened 2024-04-10
Add support hanami router

AlexeyMatskevich:
As part of this MR I added the ability to analyse the hanami router for proper documentation generation and to support generation of tags and summary similar to how it happens in rails.

I can also highlight the following changes:
1. Framework-specific code is highlighted in extractor.
2. Extractors are loaded only if there are corresponding dependencies in bundler.
3. Framework-specific features such as Active Support for rails or dry for hanami can be used within extractors.
4. Attempting to lay the architecture for creating similar extractors for any other framework.

Less significant changes:
1. I tried to make openapi documentation for rails and hanami as identical as possible, so that it can be compared via diff and see where the generation differs. 
2. For hanami, a dry inflector is used to convert the strings

Now, it seems that in almost all cases the extractors for hanami and rails work identically, but I could be mistaken.


lib/rspec/openapi/extractors/hanami.rb
```diff
@@ -0,0 +1,118 @@
+# frozen_string_literal: true
+
+require 'dry/inflector'
+require 'hanami'
+
+# Hanami::Router::Inspector original code
```
exoego, line 6: What does this mean?
Did you copy the code from Hanami?

lib/rspec/openapi/minitest_hooks.rb
```diff
@@ -13,11 +13,27 @@ def run(*args)
...
+        RSpec::OpenAPI::Extractors::Rails
+      elsif names.include?('hanami') && defined?(Hanami) &&
+            Hanami.respond_to?(:app) && Hanami.app?
+        RSpec::OpenAPI::Extractors::Hanami
+        # elsif defined?(Roda)
+        #   some Roda extractor
+      else
+        RSpec::OpenAPI::Extractors::Rack
+      end
+    end
```
exoego, line 36: Can we extract `find_extractor` from `minitest_hooks.rb` and `rspec_hooks.rb` and deduplicate?

spec/apps/hanami/doc/openapi.yaml
```diff
@@ -212,16 +236,11 @@ paths:
               example:
                 items:
                 - secrets
-        '401':
```
exoego, line 215: Hmm, why `401` responses are gone?
It looks like a regression.

lib/rspec/openapi.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'bundler/setup'
```
exoego, line 3: Is `bundler` gem always included in runtime?
If not, it should be added to `add_dependency` section:
https://github.com/exoego/rspec-openapi/blob/cdda8fc479618f3c44728f84c67eeb938df99aef/rspec-openapi.gemspec#L31-L33



Since `bundler` is not so small (400KB+ since 2.4.3), it would be preferable not to include extra dependency by using Ruby's native mechanism like `defined?(Rails)` or `defined?(Hanami)`.

lib/rspec/openapi/extractors/hanami.rb
exoego: I assume it is from https://github.com/hanami/router/blob/97f75b8529574bd4ff23165460e82a6587bc323c/lib/hanami/router/inspector.rb#L13

They are MIT License now https://github.com/hanami/router/blob/97f75b8529574bd4ff23165460e82a6587bc323c/LICENSE.md so the code does not conflict with here.

AlexeyMatskevich: Yes, I took their class for an example, and am referring to the original class, perhaps that's not a very valid comment.

exoego: Could you replace the comment with a link to `https://github.com/hanami/router/blob/97f75b8529574bd4ff23165460e82a6587bc323c/lib/hanami/router/inspector.rb#L13`?
It might help when Hanami introduces breaking change to `Inspector `

spec/apps/hanami/doc/openapi.yaml
AlexeyMatskevich: I have researched this issue, missed it unfortunately earlier. I think it's not a regression, it's just that this section of documentation in rails is manually added, maybe it's part of testing that checks that manually added documentation won't be removed automatically.

When I generated the hanami documentation I deleted the whole file and re-generated it, I will add this part manually if I understand the reason for the existence of this part correctly.

lib/rspec/openapi.rb
AlexeyMatskevich: I tried not using bundler and using the standard methods you suggested, however at this point it seems due to the loading order, these constants don't exist yet.
If i try this:
```ruby
module SharedHooks
  def self.find_extractor
    if defined?(Rails) && Rails.respond_to?(:application) && Rails.application
      require 'rspec/openapi/extractors/rails'

      RSpec::OpenAPI::Extractors::Rails
    elsif defined?(Hanami) && Hanami.respond_to?(:app) && Hanami.app?
      require 'rspec/openapi/extractors/hanami'

      RSpec::OpenAPI::Extractors::Hanami
      # elsif defined?(Roda)
      #   some Roda extractor
    else
      require 'rspec/openapi/extractors/rack'

      RSpec::OpenAPI::Extractors::Rack
    end
  end
end
```
it won't work for hanami because of the way I use to throw the inspector into the router.

Would be glad for help in this aspect

exoego: How about 

```ruby
begin
   require 'hanami'
   require 'rspec/openapi/extractors/hanami'
rescue
   puts 'Hanami not detected'
end

begin
   require 'rails'
   require 'rspec/openapi/extractors/rails'
rescue
   puts 'Rails not detected'
end
```

exoego (approved):
LGTM 
Thanks

exoego:
This was released in v0.17.0

PR #210, opened 2024-04-12
Generate multiple examples

AlexeyMatskevich:
I tried to do the generation of a few examples as [described here](https://github.com/exoego/rspec-openapi/issues/177). I would like to get a review of my approach before I continue working on this feature.

The very design of this feature is also questionable, perhaps it makes sense to use the feature in roughly this way for simpler cases:
```ruby
describe '#index', openapi: { enable_examples: true } do
    context 'when foo', openapi: { example_name: :foo }
    context 'when bar', openapi: { example_name: :bar }
end
```
In other cases I am concerned about the definition of unique examples, do I need to remove non-unique examples, or should I keep all examples? Maybe both options are needed for flexibility 🤔

### DIFF
--- a/lib/rspec/openapi.rb
+++ b/lib/rspec/openapi.rb
@@ -13,6 +13,7 @@ require 'rspec/openapi/schema_sorter'
 require 'rspec/openapi/key_transformer'
 require 'rspec/openapi/shared_hooks'
 require 'rspec/openapi/extractors'
+require 'rspec/openapi/extractors/shared_extractor'
 require 'rspec/openapi/extractors/rack'
 
 begin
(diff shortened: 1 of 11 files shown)


AlexeyMatskevich:
@exoego take a look, please

exoego:
I am fine with the `example_name` approach.

How does this handle when some test cases have `example_name` but other cases don't for an endpoint?


> In other cases I am concerned about the definition of unique examples, do I need to remove non-unique examples, or should I keep all examples? Maybe both options are needed for flexibility 🤔

I don't get the concern.
When do users specify same `example_name` to multiple test cases?

rhiroe:
@AlexeyMatskevich In my personal opinion,

> Then it's a flexibility issue where we want to throw all sample responses into examles or just certain ones

For the ones we don't want to include, I think adding `openapi: false` metadata would be a good approach.

> and we usually use factory_bot and from run to run the data is different and that's a problem.

Our team solves this by setting a fixed seed value when the OPENAPI environment variable is present.

```rb
RSpec.configure do |config|
  config.seed = 1 if ENV['OPENAPI']
```

Spence1115:
@AlexeyMatskevich any more thoughts on progressing with this?

AlexeyMatskevich:
Okay, I'm closing this MR because I created a new one where I implemented everything with a different design and approach.

PR #212, opened 2024-04-16
Don't rescue when `require 'rspec/openapi/*'` raises `LoadError`

tricknotes:
It is useful for debug.
If `require 'rspec/openapi/*'` raises `LoadError`, previous code hides original cause.

### DIFF
--- a/lib/rspec/openapi.rb
+++ b/lib/rspec/openapi.rb
@@ -17,16 +17,18 @@ require 'rspec/openapi/extractors/rack'
 
 begin
   require 'hanami'
-  require 'rspec/openapi/extractors/hanami'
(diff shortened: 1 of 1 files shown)


exoego (commented):
Why an extractor is needed when its corresponding framework is not loaded?

tricknotes:
An extractor is loaded after framework is loaded because it is **only** required in `else` case.

Example code is here:
``` ruby
begin
  raise
rescue => e
  puts 'raise'
else
  puts 'else'
end
#=> raise
```

exoego (approved):
Ah, I see.
I took `else` for `ensure`.
`else` section is executed only when no error raised.

PR #213, opened 2024-04-16
Add support mounted rack app for rails and hanami

AlexeyMatskevich:
I noticed that in cases where a rack application is mounted to main application, like pure rack or roda, in these cases for both hanami and rails, the routing recognition does not work correctly.

My suggestion in such cases is to use rack extractor instead of the main extractor for the framework.


AlexeyMatskevich:
It looks like the tests are crashing because of the difference in generation in spec/rspec/rails_spec.rb and spec/minitest/rails_spec.rb. 
I don't see what the difference is yet

exoego (approved):
Thanks. 
I fixed `rails_test.rb` for minitest.

PR #216, opened 2024-04-17
fix: warn when Hanami or Rails are not detected

bf4:
I started noticing "Hanami not detected" lately.  I'm not sure exactly what to recommend here, perhaps only output anything when `ENV["CI"] == "true"` or something like that.. but I think preferring warn -> stderr over puts -> stdout is a good start to lower the shoutiness of it.


exoego:
Thanks.
Superseded by #217

PR #230, opened 2024-04-23
Get transforms to Post if params given

rodrieiz:
I have the same issue that is mentioned [here](https://github.com/rspec/rspec-rails/issues/2502).
I think this is a possible solution that worked for me.

### DIFF
--- a/lib/rspec/openapi/extractors/hanami.rb
+++ b/lib/rspec/openapi/extractors/hanami.rb
@@ -55,7 +55,7 @@ class << RSpec::OpenAPI::Extractors::Hanami = Object.new
   # @param [RSpec::Core::Example] example
   # @return Array
   def request_attributes(request, example)
-    route = Hanami.app.router.recognize(request.path, method: request.method)
+    route = Hanami.app.router.recognize(request.path, method: request.request_method)
 
     return RSpec::OpenAPI::Extractors::Rack.request_attributes(request, example) unless route.routable?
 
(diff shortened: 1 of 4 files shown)


exoego:
I don't get the issue.
Why you need to specify `as: :json` for `get` HTTP method?

exoego:
As [the](https://github.com/rspec/rspec-rails/issues/2502) said,  

```ruby
      get "/users/1/billing_addresses",
          as: :json,
          headers: {},
          params: { some_param: {} } 
```

fails with `No route matched for POST /users/1/billing_addresses` even with this PR.

Please add a failing test first then change codes to fix the test.

rodrieiz:
> I don't get the issue. Why you need to specify `as: :json` for `get` HTTP method?

Looking for an answer to your question about why I have to use as: :json, I think that's the issue. I shouldn't use it in GET requests. But since it was like that throughout the entire project, I kept using it. Thanks.

rodrieiz:
> As [the rspec-rails issue](https://github.com/rspec/rspec-rails/issues/2502) said,
> 
> ```ruby
>       get "/users/1/billing_addresses",
>           as: :json,
>           headers: {},
>           params: { some_param: {} } 
> ```
> 
> actually fails with `No route matched for POST /users/1/billing_addresses`, but it still fails even with this PR.
> 
> Please add a failing test first then change codes to fix the test.

In my case, the tests don't fail neither before nor after the PR. It's just that before, the documentation documented the endpoint as POST instead of GET. But as I mentioned in the previous comment, I think the issue lies in using `as: :json`

PR #231, opened 2024-04-30
Use Rack::MockRequest for correct router recognize on hanami app

AlexeyMatskevich:
This should fix https://github.com/hanami/router/issues/271

### DIFF
--- a/lib/rspec/openapi/extractors.rb
+++ b/lib/rspec/openapi/extractors.rb
@@ -2,11 +2,4 @@
 
 # Create namespace
 module RSpec::OpenAPI::Extractors
-  # @param [String, Symbol] path_parameter
-  # @return [Integer, nil]
-  def self.number_or_nil(path_parameter)
-    Integer(path_parameter.to_s || '')
-  rescue ArgumentError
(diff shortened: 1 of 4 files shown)


exoego:
Thanks.
This fixes https://github.com/exoego/rspec-openapi/issues/227

Do you think https://github.com/hanami/router/issues/271 is not an issue in Hanami, but in rspec-openapi?

AlexeyMatskevich:
I talked about a similar problem on the [hanami forum](https://discourse.hanamirb.org/t/router-recognize/961/2) (I advise you to write there by the way, because they almost never respond in github and other places) - apparently, this is not an issue, but an undocumented nuance of operation.

They have examples of their work in their documentation https://www.rubydoc.info/gems/hanami-router/Hanami%2FRouter:recognize. And there are examples of work with and without Rack::MockRequest, but there is no description of such nuances that without Rack::MockRequest recognition does not always work validly.

So I think it's just a Hanami documentation problem.

AlexeyMatskevich:
@exoego Is there anything else I need to do in this MR?

PR #233, opened 2024-05-06
Fix array in query params

lan-tripla:
issue: https://stackoverflow.com/questions/68452127/pass-array-as-query-parameter-in-swagger-openapi-3-0

### DIFF
--- a/lib/rspec/openapi/schema_builder.rb
+++ b/lib/rspec/openapi/schema_builder.rb
@@ -118,6 +118,8 @@ class << RSpec::OpenAPI::SchemaBuilder = Object.new
     if value.is_a?(Hash) && (value_keys = value.keys).size == 1
       value_key = value_keys.first
       build_parameter_name("#{key}[#{value_key}]", value[value_key])
+    elsif value.is_a?(Array)
+      "#{key}[]"
     else
       key
     end
--- a/spec/apps/rails/doc/openapi.json
+++ b/spec/apps/rails/doc/openapi.json
@@ -592,6 +592,20 @@
             }
           },
           {
+            "name": "names[]",
+            "in": "query",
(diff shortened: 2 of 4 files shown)


exoego:
Could you add test for this

lan-tripla:
> Could you add test for this

updated test, could you check?

exoego:
@lan-tripla 
Sorry for very late response.

I am skeptic to this change, since it is not mandatory to name array-type query parameters like `foo[]` in [OpenAPI 3.0](https://swagger.io/docs/specification/v3_0/describing-parameters/#query-parameters), IIUC.
Even an OpenAPI 3.0 example from https://editor.swagger.io/ shows the example of name without `[]`:
 
```yaml
  /pet/findByStatus:
    get:
      tags:
        - pet
      summary: Finds Pets by status
      description: Multiple status values can be provided with comma separated strings
      operationId: findPetsByStatus
      parameters:
        - name: status
          in: query
          description: Status values that need to be considered for filter
          required: false
          explode: true
          schema:
            type: string
            default: available
            enum:
              - available
              - pending
              - sold
```

The stackoverflow issue you cited appears very specific issue to node-express's implementation for OpenAPI support.

I am fine to accept this change if a new option is added to name query params with `[]` and the option defaults to disabled for compatibility

exoego:
Closing since there's been no response to the request and it has been sitting idle for a long time.

For the record: `[]` suffix on array query parameter names is not required by the OpenAPI 3.0 spec.
The reported failure is specific to certain tooling.
The example at https://editor.swagger.io/ also shows array-typed query parameters declared without `[]`.

A follow-up PR that adds this as a new opt-in option (defaulting to disabled for backward compatibility) would be very welcome.

PR #236, opened 2024-09-11
Makes header & request params optional and handle empty files gracefully

muhammadnawzad:
Hello good people, I hope all is well.

This PR introduces three minor changes to the `rspec-openapi` gem:

1. **Header params can now be optional and by default required:**
  In the previous implementation, all header params were marked as required (`required: true`). Now, header params are optional, and by default required. The logic has been updated to check the `optional_headers` field, and any headers not present in that list are considered required (`record.optional_headers.exclude?(key)`)

2. **Requets body params can now be optional and by default required:**
  In the previous implementation, all request body params were marked as required from the `enrich_with_required_keys` method, (`obj[:required] = obj[:properties]&.keys || []`). Now, request body params are optional, and by default required. The logic has been updated to check the `optional_request_params` field in the record, and any params not present in that list are considered required (`(obj[:properties]&.keys - optional_request_params) || []`)

3. **Gracefully handle empty files during YAML/JSON parsing:**
  The previous implementation would raise errors when attempting to load an empty YAML/JSON file. The update adds a check to ensure that if the content is nil (i.e., the file is empty), an empty hash {} is returned, avoiding any errors. _(Check the provided error at the bottom)._

These changes improve flexibility in defining required headers and prevent errors when dealing with empty files. Please note that I have tested these on a **Rails API**, provided changes should be tested in other frameworks.

P.S. Automated unit test is not provided since the changes doesn't modify the intended behavior of these methods by default current tests should cover these as well.

Raised Error mentioned in 3:
```
Failure/Error: base.replace(RSpec::OpenAPI::KeyTransformer.symbolize(base))

NoMethodError:
  undefined method `replace' for nil
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,6 +337,8 @@ Some examples' attributes can be overwritten via RSpec metadata options. Example
     description: 'list all posts ordered by pub_date',
     tags: %w[v1 posts],
     required_request_params: %w[limit],
+    optional_request_params: %w[data optional_address], # request body parameters that will NOT be set as required
+    optional_headers: %w[Accept-Languages rowsPerPage], # request header params that can be optional
     security: [{"MyToken" => []}],
   } do
     # ...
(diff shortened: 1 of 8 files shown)


exoego:
@muhammadnawzad Sorry for very long response.

I am happy to accept this PR if tests are added for new option enabled.

exoego:
Thanks for the PR.

- Point 1 was fixed in #338 
- Point 2 was already implemented without metadata like `optional_request_params` sometimes 
- Point 3 was fixed in #339

Issue #171, opened 2024-01-31
Evaluate proc for RSpec::OpenAPI.title


alexpapworth:
My use case is generating different openapi files from different RSpec request files. The `RSpec::OpenAPI.path` proc is great to export YAML to different files, but I was surprised to see I couldn't do the same with `RSpec::OpenAPI.title`

Unfortunately the following snippet generates: `title: !ruby/object:Proc {}`

```ruby
RSpec::OpenAPI.title = -> (example) {
  return "Customer API" if example.file_path == "./spec/requests/customer_api_spec.rb"
  return "Reseller API" if example.file_path == "./spec/requests/resellers_api_spec.rb"

  "API"
}
```

exoego:
How about https://github.com/exoego/rspec-openapi/pull/162

alexpapworth:
Ah interesting. My usecase is slightly different. I would favour a proc on `RSpec::OpenAPI.title` in a single config file rather than creating the multiple config.

josefggx:
@exoego @alexpapworth, Would it be ideal to assign the title based on the path of the example, or on the file path of the resulting schema? Because if the example is used, I think there could be more titles than schema files, and titles would end up being overwritten if not managed properly. I can provide a solution for both, let me know which would be the preferred configuration:

```ruby
# Option A - Resulting schema file path:
RSpec::OpenAPI.title = -> (path) {
  case path
  when %r[doc/openapi/v1.yaml] then 'API v1 Documentation'
  when %r[doc/openapi/v2.yaml] then 'API v2 Documentation'
  else 'OpenAPI Documentation'
  end
}

# Option B - Example file path:
RSpec::OpenAPI.title = -> (example) {
  case example.file_path
  when %r[spec/requests/api/v1/] then 'API v1 Documentation'
  when %r[spec/requests/api/v2/] then 'API v2 Documentation'
  else 'OpenAPI Documentation'
  end
}
```

alexpapworth:
Ooh interesting. I'm leaning towards Option B, because then the lambda will take the same parameters as `RSpec::OpenAPI.path = ` currently does.

exoego:
This was fixed in #240 and published in v0.18.4

PR #240, opened 2024-10-25
Evaluate proc for RSpec::OpenAPI.title

josefggx:
### Description:
Solution to #171. With the additional logic you can make the title dynamic using the same approach used with the path, and maintain backward compatibility. This allows the generation of individual schema files, each with its own title.

### Example:
It's possible to assign titles in 2 ways:
```ruby
# Change the default title of the generated schema
RSpec::OpenAPI.title = 'OpenAPI Documentation'

# Or generate individual titles for your partial schema files, given an RSpec example
RSpec::OpenAPI.title = -> (example) {
  case example.file_path
  when %r[spec/requests/api/v1/] then 'API v1 Documentation'
  when %r[spec/requests/api/v2/] then 'API v2 Documentation'
  else 'OpenAPI Documentation'
  end
}
```

I updated the documentation with this information. I left a [comment](https://github.com/exoego/rspec-openapi/issues/171#issuecomment-2429550727) with an alternative solution in case it's preferred

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,17 @@ RSpec::OpenAPI.path = -> (example) {
   end
 }
 
+# Change the default title of the generated schema
 RSpec::OpenAPI.title = 'OpenAPI Documentation'
+
+# Or generate individual titles for your partial schema files, given an RSpec example
+RSpec::OpenAPI.title = -> (example) {
+  case example.file_path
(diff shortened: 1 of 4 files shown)


exoego (approved):

exoego:
@josefggx  Thanks for the PR. This was released in https://rubygems.org/gems/rspec-openapi/versions/0.18.4

PR #243, opened 2025-02-03
Fix semantic error caused by requestBody in deletes

dantherussell:
While testing some generated docs which included a `DELETE` operation, the [Swagger editor](https://editor.swagger.io/) gave me the following error:

```
Semantic error at paths./views/{id}.delete.requestBody
DELETE operations cannot have a requestBody.
```

When I looked into it, I found out that in [v3.0.3 of the OpenAPI Spec](https://spec.openapis.org/oas/v3.0.3#fixed-fields-7) (which is what this gem currently generates), a `requestBody` in a `DELETE` operation is not allowed, as the HTTP 1.1 specification [Section 4.3.1](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.1) does not have explicitly defined semantics for request bodies in `DELETE`.

This has changed in [v3.1.0](https://spec.openapis.org/oas/v3.1.0#fixed-fields-7), where for such operations, `requestBody` is permitted but does not have well-defined semantics and, as per the specification, should be avoided if possible.

The `schema_builder` was already using a nil `requestBody` for `GET` requests, so I have modified it slightly to do the same for `DELETE`.

### DIFF
--- a/lib/rspec/openapi/schema_builder.rb
+++ b/lib/rspec/openapi/schema_builder.rb
@@ -36,7 +36,7 @@ class << RSpec::OpenAPI::SchemaBuilder = Object.new
             security: record.security,
             deprecated: record.deprecated ? true : nil,
             parameters: build_parameters(record),
-            requestBody: http_method == 'get' ? nil : build_request_body(record),
+            requestBody: include_nil_request_body?(http_method) ? nil : build_request_body(record),
             responses: {
               record.status.to_s => response,
             },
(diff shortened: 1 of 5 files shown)


dantherussell:
FYI @exoego that logger issue is something to do with a gem dependancy changing, I can fix it in this PR if you'd like?

exoego:
@dantherussell Nice but can you create a separate PR?

dantherussell:
> @dantherussell Nice but can you create a separate PR?

Sure - https://github.com/exoego/rspec-openapi/pull/244

exoego (approved):
LGTM
Will look Hanami issue later

exoego:
@dantherussell Thanks for the PRs. This was released in https://rubygems.org/gems/rspec-openapi/versions/0.18.4

PR #244, opened 2025-02-04
Fix rspec issues caused by changes to concurrent-ruby

dantherussell:
As per https://stackoverflow.com/a/79361034, concurrent-ruby v1.3.5 has removed the dependency on logger, causing issues with the activesupport gem.


dantherussell:
No idea why the Hanami tests are failing

exoego:
Thanks.
At least Ruby 2.7 now passes.
Will look Hanami issue later

PR #256, opened 2025-06-12
Fix psych  disallowedclass error in ruby 3.1

ipepe:
Fixes https://github.com/exoego/rspec-openapi/issues/249 by adding `Date` to `permitted_classes` of `YAML.safe_load` in `RSpec::OpenAPI::SchemaFile#read`.

# Changes

- allow Date and Time when reading YAML schema files
- add regression test for SchemaFile

### DIFF
--- a/lib/rspec/openapi/schema_file.rb
+++ b/lib/rspec/openapi/schema_file.rb
@@ -3,6 +3,7 @@
 require 'fileutils'
 require 'yaml'
 require 'json'
+require 'date'
 
 # TODO: Support JSON
 class RSpec::OpenAPI::SchemaFile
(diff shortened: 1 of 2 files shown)


lib/rspec/openapi/schema_file.rb
```diff
@@ -3,6 +3,8 @@
 require 'fileutils'
 require 'yaml'
 require 'json'
+require 'date'
+require 'time'
```
rhiroe, line 7: [nits]
`require 'yaml'` also loads `Date`. `Time` is available without requiring anything explicitly.

spec/rspec/schema_file_spec.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe RSpec::OpenAPI::SchemaFile do
+  describe '#read' do
```
rhiroe, line 6: [imho]
How about adding unquoted dates or times to the OpenAPI schema in `spec/apps/` instead of writing tests for the private method?

What should be tested is not whether unquoted dates or times are loaded, but whether the resulting OpenAPI schema output is correct—and whether it can be successfully loaded.

lib/rspec/openapi/schema_file.rb
exoego: https://github.com/exoego/rspec-openapi/pull/261

PR #267, opened 2025-11-15
feat: evaluate array item nullability across the array

### LINKED ISSUE #258: bug: types not being evaluated correctly
The types seem to only be evaluated against the first object in the array

```json
"paths": {
    "/my_endpoint": {
      "get": {
        "summary": "index",
        "tags": [
          "MyEndpoint"
        ],
        "responses": {
          "200": {
            "description": "returns data",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "properties": {
                    "message": {
                      "type": "string"
                    },
                    "data": {
                      "type": "array",
                      "items": {
                        "type": "object",
                        "properties": {
                          "name": {
                            "type": "string"
                          },
                          "description": {
                            "nullable": true
                          }
                        },
                        "required": [
                          "name",
                          "description"
                        ]
                      }
                    },
                    "result_code": {
                      "type": "string"
                    }
                  },
                  "required": [
                    "message",
                    "data",
                    "result_code"
                  ]
                },
                "example": {
                  "message": "Success",
                  "data": [
                    {
                      "name": "Marketing",
                      "description": null
                    },
                    {
                      "name": "R&D",
                      "description": "Research and Development"
                    }
                  ],
                  "result_code": "success"
                }
              }
            }
          }
        }
      }
    }
  }
```

You can see from the example produced by the spec that `description` is a `string`, but is also `nullable`. However, `rspec-openapi` has only logged that it is `nullable`, not that its a `string`. If I ensure the data is displayed with the 2nd entry first, then it picks up that its a `string`, but not that its `nullable`.

Spence1115:
## Summary

When evaluating an array of hashes, currently only the first entry in the array is used to check for values. This means we never see if something is nullable and only present in _some_ entries in the example

This PR adds methods to iterate through the array and assess nullability based on each item in the array. For example, a response body like:
```
{
    "users" => [
    {
        "label" => "Jane Doe",
        "value" => "jane_doe"
    },
    {
        "label" => nil,
        "value" => "unknown"
    }
    ],
}
```

Will produce properties like:
```
{
    "label" => {
        type: "string",
        nullable: true
    },
    "value" => {
        type: "string"
    }
}
```

Additionally, any hashes with keys that are not present in other hashes should be considered `nullable` and not `required`

```
{
    "users" => [
        {
            "label" => "Jane Doe",
            "value" => "jane_doe"
        },
        {
            "label" => nil,
            "value" => "unknown",
            "invited" => true
        }
    ],
}
```

Produces:

```
schema: {
  type: "object",
  properties: {
    "users" => {
      type: "array",
      items: {
        type: "object",
        properties: {
          "label" => {
            type: "string",
            nullable: true
          },
          "value" => {
            type: "string"
          },
          "invited" => {
            type: "boolean",
            nullable: true
          }
        },
        required: ["label", "value"]
      }
    }
  },
  required: ["users"]
},
```

Resolves #258

### DIFF
--- a/spec/requests/hanami_spec.rb
+++ b/spec/requests/hanami_spec.rb
@@ -311,21 +311,21 @@ RSpec.describe 'non-numeric path parameter', type: :request do
 end
 
 RSpec.describe 'Array of hashes', type: :request do
-  describe "with nullable keys" do
+  describe 'with nullable keys' do
     it 'returns some content' do
       get '/array_hashes/nullable'
       expect(last_response.status).to eq(200)
     end
   end
 
-  describe "with non-nullable keys" do
+  describe 'with non-nullable keys' do
     it 'returns some content' do
       get '/array_hashes/non_nullable'
       expect(last_response.status).to eq(200)
     end
   end
 
-  describe "with nullable keys" do
+  describe 'with nullable keys' do
     it 'returns some content' do
       get '/array_hashes/nullable'
       expect(last_response.status).to eq(200)


Spence1115:
> Integration test is more preferred than unit with mock/double in this repo, like done in #226 It allows drastic changes of internal classes without maintaining unit tests.

@exoego changed to integration tests

lib/rspec/openapi/schema_builder.rb
```diff
@@ -243,4 +243,78 @@ def normalize_content_type(content_type)
...
+          prop1 = merged[:properties][key]
+          merged[:properties][key] = merge_property_schemas(prop1, prop2)
+        else
+          merged[:properties][key] = make_property_nullable(prop2)
+        end
+      end
+
+      schema1[:properties].each do |key, prop1|
+        unless schema2[:properties][key]
+          merged[:properties][key] = make_property_nullable(prop1)
```
exoego: not covered according to https://app.codecov.io/gh/exoego/rspec-openapi/pull/267

lib/rspec/openapi/schema_builder.rb
```diff
@@ -243,4 +243,78 @@ def normalize_content_type(content_type)
...
+    merged = prop1.dup
+
+    # If either property is nullable, the merged property should be nullable
+    if prop2[:nullable] && !prop1[:nullable]
+      merged[:nullable] = true
+    end
+
+    # If both are objects, recursively merge their properties
+    if prop1[:type] == 'object' && prop2[:type] == 'object'
+      merged = merge_object_schemas(prop1, prop2)
```
exoego: not covered according to https://app.codecov.io/gh/exoego/rspec-openapi/pull/267

Spence1115: Should now be covered

lib/rspec/openapi/schema_builder.rb
Spence1115: Should now be covered

PR #293, opened 2025-12-24
Support multiple examples via openapi: { example_mode: :multiple }

AlexeyMatskevich:
## Summary

Add support for generating multiple named examples in OpenAPI specification via the new `example_mode` configuration option.

---

## Motivation

OpenAPI's `examples` (plural) feature allows multiple named response variants per endpoint. This PR enables rspec-openapi to generate these named examples automatically from RSpec tests.

Each test becomes a named example where:
- The key is derived from test description (e.g., `returns_orders_for_admin`)
- The summary preserves the human-readable test description
- The value contains the actual response

This makes different behaviors and edge cases visible directly in API documentation. Frontend developers can see multiple scenarios for the same endpoint, with meaningful names from your RSpec specs.

---

### New Configuration Option: `example_mode`

Available values:
- `:single` (default) - generates single `example` field (backward compatible)
- `:multiple` - generates named `examples` with test descriptions as keys
- `:none` - generates only schema, no examples

### New Global Settings

```ruby
# Customize example name generation (used for multiple examples)
RSpec::OpenAPI.example_name_builder = ->(example) { example.description }

# Disable generating example summaries for `examples`
RSpec::OpenAPI.enable_example_summary = false
```

### Usage via RSpec Metadata

```ruby
describe '#index', openapi: { example_mode: :multiple } do
  it 'with pagination' do
    get '/tables', params: { page: 1, per: 10 }
  end

  it 'with filter' do
    get '/tables', params: { filter: { name: 'test' } }
  end
end
```

Generated output:
```yaml
responses:
  '200':
    content:
      application/json:
        schema: { ... }
        examples:
          with_pagination:
            summary: "with pagination"
            value: { ... }
          with_filter:
            summary: "with filter"
            value: { ... }
```

---

### 1. Basic Modes
| Mode | Behavior | Use Case |
|------|----------|----------|
| `:single` | Generates `example` field | Default behavior, single representative example |
| `:multiple` | Generates `examples` with named keys | Multiple scenarios for same endpoint |
| `:none` | Schema only, no examples | Edge case tests, internal endpoints |

### 2. Mode Inheritance
Mode is inherited through nested RSpec contexts:
```ruby
describe 'API', openapi: { example_mode: :multiple } do
  describe '#index' do
    it 'first example' do ... end   # inherits :multiple
    it 'second example' do ... end  # inherits :multiple
  end
end
```

### 3. Mode Override
Child contexts can override parent mode:
```ruby
describe 'API', openapi: { example_mode: :multiple } do
  describe '#show', openapi: { example_mode: :single } do
    it 'returns item' do ... end  # uses :single
  end
end
```

### 4. Mixed Mode Merging
When different tests target the same endpoint with different modes, the merger auto-converts to `examples`:
```ruby
# File A - default :single mode
it 'returns users' do
  get '/users'
end

# File B - :multiple mode
describe 'GET /users', openapi: { example_mode: :multiple } do
  it 'with admin privileges' do
    get '/users', headers: { 'X-Admin': 'true' }
  end
end
```
Result: Both examples merged into `examples` field.

### 5. Global Example Disabled
When `RSpec::OpenAPI.enable_example = false`, all modes respect this and generate schema-only output.

### 6. Example Key Collision
If multiple examples resolve to the same key, the last one wins (overwrites). This is documented behavior.

---

## Backward Compatibility

✅ **Fully backward compatible**

- Default mode is `:single`, which produces the same output as before
- No breaking changes to existing configuration options
- Existing specs continue to work without modification

---

## Related

Rebased from original PR #210 to v0.21.5

### DIFF
--- a/spec/apps/hanami/config/routes.rb
+++ b/spec/apps/hanami/config/routes.rb
@@ -59,6 +59,12 @@ module HanamiTest
     get '/example_summary_disabled', to: ->(_env) { [200, { 'Content-Type' => 'application/json' }, ['{"data":"no_summary"}']] }
     get '/empty_example_name', to: ->(_env) { [200, { 'Content-Type' => 'application/json' }, ['{"data":"empty_name"}']] }
 
+    # Test route for nested arrays (key_transformer coverage)
+    get '/nested_arrays_test', to: ->(_env) { [200, { 'Content-Type' => 'application/json' }, ['{"items":[{"name":"first","tags":["a","b","c"]},{"name":"second","tags":["x","y","z"]}],"matrix":[[1,2],[3,4]]}']] }
+
+    # Test route for invalid example_mode error handling
+    get '/invalid_example_mode', to: ->(_env) { [200, { 'Content-Type' => 'application/json' }, ['{"status":"ok"}']] }
+
     slice :my_engine, at: '/my_engine' do
       get '/test', to: ->(_env) { [200, { 'Content-Type' => 'text/plain' }, ['ANOTHER TEST']] }
       get '/eng/example', to: 'eng.example'
--- a/spec/apps/hanami/doc/openapi.json
+++ b/spec/apps/hanami/doc/openapi.json
@@ -1771,6 +1771,90 @@
         }
       }
     },
+    "/nested_arrays_test": {
+      "get": {
+        "summary": "GET /nested_arrays_test",
+        "tags": [],
+        "responses": {
+          "200": {
+            "description": "returns nested arrays",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "properties": {
+                    "items": {
+                      "type": "array",
+                      "items": {
+                        "type": "object",
+                        "properties": {
+                          "name": {
+                            "type": "string"
+                          },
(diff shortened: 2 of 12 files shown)


lib/rspec/openapi/extractors/shared_extractor.rb
```diff
@@ -0,0 +1,61 @@
...
+  def self.normalize_example_mode(value, example = nil)
+    return :single if value.nil?
+
+    raise ArgumentError, example_mode_error(value, example) unless value.is_a?(String) || value.is_a?(Symbol)
+
+    mode = value.to_s.strip.downcase.to_sym
+    return mode if VALID_EXAMPLE_MODES.include?(mode)
+
+    raise ArgumentError, example_mode_error(value, example)
+  end
```
AlexeyMatskevich, line 54: @exoego I cannot cover with integration tests this code, because it raise exception in after(:each) block, and becouse this exception just for better developer experience.

exoego: It is OK.

exoego (approved):
Great work.
Thank you very much

PR #294, opened 2025-12-24
fix: safely extract request/response when let() overrides exist

AlexeyMatskevich:
## Summary
- Fixes OpenAPI extraction when users define `let(:request)` or `let(:response)` in their specs
- Prioritizes `integration_session` if available for request/response extraction
- Falls back to detecting RSpec memoized helpers and calling `super_method` to get actual objects

## Test plan
- [x] Added integration test that defines `let(:request)` and `let(:response)` overrides
- [x] Test verifies OpenAPI record is still generated with correct status code

Related issue https://github.com/exoego/rspec-openapi/issues/251


exoego:
Thanks.
Merged via https://github.com/exoego/rspec-openapi/pull/334

PR #305, opened 2026-01-11
fix: correct handling for nested query parameters

### LINKED ISSUE #306: bug: incorrect handling of nested query params
Nested query params are being doubled up. For example, in the existing spec
```ruby
      it 'with deep query parameters' do
        get '/tables', params: { filter: { 'name' => 'Example Table' } }, headers: { authorization: 'k0kubun' }
      end
```

We would expect OpenAPI to output a schema like:
```yaml
      - name: filter[name]
        in: query
        required: false
        schema:
          type: string
        example:
          name: Example Table
```
Or
```yaml
      - name: filter
        in: query
        required: false
        schema:
          type: object
          properties:
            name:
              type: string
          required:
          - name
        example:
          name: Example Table
```

Instead, we get:

```yaml
      - name: filter[name]
        in: query
        required: false
        schema:
          type: object
          properties:
            name:
              type: string
          required:
          - name
        example:
          name: Example Table
```

Which describes an object lke
```
filter: {
  name: {
    name: "Example Table"
  }
}
```

Spence1115:
Nested filter queries are handled incorrectly, leading to double nesting:

**Before**
```ruby
      it 'with deep query parameters' do
        get '/tables', params: { filter: { 'name' => 'Example Table' } }, headers: { authorization: 'k0kubun' }
      end
```
```yaml
      - name: filter[name]
        in: query
        required: false
        schema:
          type: object
          properties:
            name:
              type: string
          required:
          - name
        example:
          name: Example Table
```

Incorrect, as this now expected a query param of `filter[name]={name:"string"}`. For example, when converted to Typescript, it comes out like:
```typescript
'filter[name]'?: {
  name: string;
};
```

**Now**
```ruby
      it 'with deep query parameters' do
        get '/tables', params: { filter: { 'name' => 'Example Table' } }, headers: { authorization: 'k0kubun' }
      end
```
```yaml
      - name: filter[name]
        in: query
        required: false
        schema:
          type: string
        example:
          name: Example Table
```
```typescript
'filter[name]'?: string;
```

An alternative approach could have been taken to change `filter` to be a full object, with `name`, and `price` as optional parameters within it, but this felt more consistent with the initial approach taken, whereas swapping to a deep object structure would be more akin to a breaking change. Happy to investigate if we'd prefer it, however.

Fixes #306

### DIFF
--- a/spec/requests/rails_spec.rb
+++ b/spec/requests/rails_spec.rb
@@ -39,7 +39,7 @@ RSpec::OpenAPI.security_schemes = {
 
 RSpec.describe 'Tables', type: :request do
   describe '#index', openapi: { summary: 'Get a list of tables' } do
-    context it 'returns a list of tables' do
+    context 'returns a list of tables' do
       it 'with flat query parameters' do
         get '/tables', params: { page: '1', per: '10' },
                        headers: { authorization: 'k0kubun', 'X-Authorization-Token': 'token' }


spec/requests/rails_spec.rb
```diff
@@ -39,7 +39,7 @@
 
 RSpec.describe 'Tables', type: :request do
   describe '#index', openapi: { summary: 'Get a list of tables' } do
-    context it 'returns a list of tables' do
+    context 'returns a list of tables' do
```
Spence1115, line 42: Fixing typo

PR #308, opened 2026-01-13
fix: allow multiple types of parameters

### LINKED ISSUE #309: bug: does not support parameters with multiple types
When the parameters to an endpoint accept multiple types, we expect OpenAPI to provide a `oneOf` with the types. However it does not, and only produces a single type.

```ruby
    it 'returns a table' do
      patch '/tables/1', headers: { authorization: 'k0kubun' }, params: { name: 'test' }
    end

    it 'returns a table via a UUID' do
      patch '/tables/abc-123', headers: { authorization: 'k0kubun' }, params: { name: 'test' }
    end
```

**Expected**
```yaml
      parameters:
      - name: id
        in: path
        required: true
        schema:
          oneOf:
          - type: integer
          - type: string
        example: abc-123
```

**Actual**
```yaml
      parameters:
      - name: id
        in: path
        required: true
        schema:
          type: integer
        example: 1
```

Spence1115:
Found an issue where a controller that takes one parameter with multiple types does not get picked up correctly. For example, a controller that takes a param `id_or_uuid` which is either an `id` (number) or `uuid` (string), only one will get chosen for OpenAPI.

This fix ensures any params are converted to `oneOf` if there are more than one types.

```ruby
    it 'returns a table' do
      patch '/tables/1', headers: { authorization: 'k0kubun' }, params: { name: 'test' }
    end

    it 'returns a table via a UUID' do
      patch '/tables/abc-123', headers: { authorization: 'k0kubun' }, params: { name: 'test' }
    end
```
```yaml
      parameters:
      - name: id
        in: path
        required: true
        schema:
          oneOf:
          - type: integer
          - type: string
        example: abc-123
```

Fixes #309

### DIFF
--- a/lib/rspec/openapi/schema_merger.rb
+++ b/lib/rspec/openapi/schema_merger.rb
@@ -65,11 +65,59 @@ class << RSpec::OpenAPI::SchemaMerger = Object.new
 
     all_parameters = all_parameters.map do |parameter|
       base_parameter = unique_base_parameters[[parameter[:name], parameter[:in]]] || {}
-      base_parameter ? base_parameter.merge(parameter) : parameter
+      if base_parameter.empty?
+        parameter
+      else
+        merge_parameter_with_schema(base_parameter, parameter)
+      end
     end
 
     all_parameters.uniq! { |param| param.slice(:name, :in) }
     base[key] = all_parameters
+  end
+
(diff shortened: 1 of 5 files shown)


Spence1115:
@exoego any chance you'd be able to review this?

exoego (approved):
LGTM

PR #319, opened 2026-03-20
fix: prevent example_mode :none from overwriting response descriptions

K4sku:
resolves #318

## Summary

- When `example_mode: :none` is set, the description from a non-documented test could overwrite a description already set by a documented test, depending on random test execution order
- Adds a `_description_skip_overwrite` internal flag (similar to existing `_example_key`/`_example_summary`) so the merger preserves existing descriptions when the incoming record is from a `:none` test
- The `:none` description is still recorded if it is the first or only test for an endpoint, keeping the `description` field valid per the OpenAPI spec

## Test plan

- [ ] Existing test suite passes (no fixture changes needed -- `:none`-only endpoints keep their descriptions)
- [ ] Verify with two tests for the same endpoint (one documented, one `:none`) that running with different random seeds produces consistent descriptions


exoego:
Thanks. Merged via https://github.com/exoego/rspec-openapi/pull/333

PR #330, opened 2026-05-07
feat: support additionalProperties via openapi metadata

### LINKED ISSUE #328: feature: override generated properties on a given endpoint
We have endpoints that return a series of strings, all of which are boolean values. Eg.

```ruby
# /api/v1/permissions

{
  "data": {
    "can_do_thing": true,
    "can_do_other_thing": false
  }
}
```

If these are static, then its all fine, but if these can vary (for example, depending on user role, or driven by a database etc), then the output is too strict and not realistic.

We have a workaround right now with something like:

```ruby
RSpec::OpenAPI.post_process_hook = ->(_path, _records, spec) do
  dynamic_field = spec.dig(
    :paths,
    :"/api/v1/permissions",
    :get, :responses, :"200", :content, :"application/json", :schema,
    :properties, :data
  )

    dynamic_field.delete(:properties)
    dynamic_field.delete(:required)
    dynamic_field[:additionalProperties] = { type: "boolean" }  # adjust type as needed
  end
end
```

However, having to do this for each endpoint is a bit arduous. It'd be nice if on the endpoint, we could do something like

```ruby
it 'checks permissions', openapi: {
  additional_properties: { 'data' => { type: 'boolean' } }
} do
  get '/api/v1/permissions'
end
```

Where the key is dot notation, similar to how enums work.

Any thoughts? Happy to handle this if you think it makes sense.

Spence1115:
## Summary

Closes #328.

Adds an `additional_properties` RSpec metadata option (and `request_additional_properties` / `response_additional_properties` variants) so endpoints that return — or accept — objects with dynamic keys produce `additionalProperties` in the generated schema instead of capturing each observed key as a fixed property.

```ruby
it 'returns a permissions map', openapi: {
  additional_properties: { 'data' => { type: 'boolean' } },
} do
  get '/api/v1/permissions'
  # Response: { "data": { "can_edit": true, "can_delete": false } }
end
```

```yaml
schema:
  type: object
  properties:
    data:
      type: object
      additionalProperties:
        type: boolean
  required:
    - data
```

## Design

Plumbed through exactly the same pipeline as the existing `enum` feature:

`SharedExtractor.attributes` → all 3 framework extractors (rails / rack / hanami) → `Record` → `SchemaBuilder#infer_additional_properties`.

In `build_property`, when the value is a `Hash` and the current dot-path matches an override, the override schema replaces what would otherwise be the captured `properties` / `required`. Schema *values* are passed through verbatim so both primitive types and `$ref` schemas are supported, as you suggested in the issue.

### Path matching

Keys are dot-notation paths (the same scheme used by `enum`). One small extension: an empty-string path `''` targets the body root, which covers the common "the whole response body is a dynamic dict" case — e.g. a memberships/permissions lookup that returns ``{"<id>" => boolean}`` directly with no wrapper. `nil.to_s == ""` makes that fall out naturally at the root of the schema-builder's recursion.

### Supported value shapes

- Primitive: `{ type: 'boolean' }`, `{ type: 'integer' }`, …
- `$ref`: `{ '$ref' => '#/components/schemas/Tag' }`
- Any other valid OpenAPI schema fragment (passed through with keys symbolized to match the rest of the builder).

### Request vs response

Following the `enum` precedent: `additional_properties` applies to both sides; `request_additional_properties` / `response_additional_properties` scope to one. The request-side override targets the request body schema, request params, etc. — wherever `build_property` recurses.

## What's covered

- Top-level field replaced with `additionalProperties` (the issue's example shape).
- Root-level dynamic dict via `''` path.
- `$ref` schema as the value.
- Request-body-only override.

All four scenarios are exercised by new request specs in `spec/requests/rails_spec.rb` against a new `DynamicKeysTestController`, and the corresponding YAML/JSON fixtures are committed so the integration test (`spec/rspec/rails_spec.rb`) verifies the full round-trip.

The schema merger already handled `additionalProperties` correctly (it preserves it across regenerations and skips `properties` / `required` where it's present), so no merger changes were needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,6 +447,99 @@ it 'creates a task', openapi: {
 end
 ```
 
+### Dynamic-key Object Support (`additionalProperties`)
+
+When an endpoint returns (or accepts) an object whose keys are not part of its
+schema — for example a permissions map `{ "can_edit": true, "can_delete": false }`
+where new permissions can appear without an API change — rspec-openapi would
+otherwise capture the test response literally and emit each observed key as a
+fixed property. You can override this with the `additional_properties` metadata,
+which replaces the captured `properties` / `required` of the matched object with
+[`additionalProperties`](https://swagger.io/docs/specification/data-models/dictionaries/).
+
+Keys are dot-notation paths (the same scheme as `enum`); the empty string `""`
+targets the body root.
+
+```rb
+it 'returns a permission map', openapi: {
+  additional_properties: { 'data' => { type: 'boolean' } },
+} do
+  get '/api/v1/permissions'
+  # Response: { "data": { "can_edit": true, "can_delete": false } }
+  expect(response.status).to eq(200)
+end
+```
+
+This generates:
+
+```yaml
+schema:
+  type: object
+  properties:
+    data:
+      type: object
+      additionalProperties:
+        type: boolean
+  required:
+    - data
+```
+
+#### Root-level dynamic keys
+
+When the entire body is a dynamic dict, use `''` as the path:
+
+```rb
+it 'lists organisation memberships', openapi: {
+  additional_properties: { '' => { type: 'boolean' } },
+} do
+  get '/api/v1/organisations/acme/check_memberships'
+  # Response: { "user-hash-a": true, "user-hash-b": false }
+  expect(response.status).to eq(200)
+end
+```
+
+```yaml
+schema:
+  type: object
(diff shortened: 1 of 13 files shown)


exoego:
I have some thoughts to enhance the feature, but this looks good itself.
I will open a subsequent PR

exoego (approved):

PR #334, opened 2026-05-08
fix: safely extract request/response when let() overrides exist

### LINKED ISSUE #251: bug: response received unexpected message :status with (no args)
**Summary**

If you use the words `request` or `response` for a variable in your tests, `rspec-openapi` will not work. I know these are preassigned words in `rspec` itself, but the use of them as variables does not break `rspec`, so I wouldn't expect it to break `rspec-openapi`.

**Reproduction**
```ruby
...
describe "#index" do
  subject { get :index }
  
  before do
    allow(SomeThirdPartyGem).to receive(:call).and_return(response)
  end

  let(:response) { double(ThirdPartyGemResponse, body: "This succeeded", code: 200) }

  it "renders the index successfully" do
    expect(subject.status).to eq(200)
  end
end
```

When just running the test, `rspec` succeeds fine. When running `openapi-rspec`:
```
\e[1mFailure/Error: \e[0mexample.run
  #<Double ThirdPartyGemResponse> received unexpected message :status with (no args)
```

exoego:
Closes #251

Fixes conflict in https://github.com/exoego/rspec-openapi/pull/294


exoego:
Thanks.
Merged via https://github.com/exoego/rspec-openapi/pull/334

PR #350, opened 2026-06-05
fix: merge heterogeneous nested-array items across outer array items

### LINKED ISSUE #351: bug: nested heterogeneous array items are not merged when shapes differ across outer array items
One of our apps generates a schema for a form-style endpoint where the same kind of polymorphism appears at two depths. Heterogeneous array item shapes merge correctly at the level of the array being merged (#267), but when an array-typed property sits *inside* the items being merged, and its shape differs **across the outer items**, the nested array doesn't get merged — it sticks on the first-encountered shape.

Given a response like:

```json
{
  "form": [
    {
      "label": "User details",
      "inputs": [
        {
          "type": "columns",
          "columns": [
            { "input": { "name": "first_name", "type": "text", "validations": { "presence": true } } }
          ]
        },
        { "type": "text", "name": "email", "validations": { "presence": true } }
      ]
    },
    {
      "label": "Billing details",
      "inputs": [
        {
          "type": "columns",
          "columns": [
            {
              "input": {
                "name": "country_code",
                "type": "dropdown",
                "validations": null,
                "answer_options": [{ "label": "United Kingdom", "value": "GB" }]
              }
            }
          ]
        }
      ]
    }
  ]
}
```

The generated schema for `form.items.properties.inputs.items.properties.columns.items.properties.input` captures only the first-encountered shape (the text input): `answer_options` is missing, `validations` is not nullable, and `required` still demands the text-only shape. The recorded response example then fails to validate against its own schema, and downstream type generators (we use Orval) emit broken TypeScript types.

Notably, the same shapes merge fine when they appear within a *single* `columns` array — the bug only triggers when the variation is spread across different outer items (`form[0]` vs `form[1]` here).

Cause: `SchemaBuilder#merge_multi_recursive` has no handling for array-typed properties. `merge_multi_array_items` has an explicit `'array'` branch that merges item variations, but in recursive-merge mode array properties fall through to `prop_variations.first.dup`, so later variations are silently dropped.

We're currently working around this with a `post_process_hook` that copies the missing properties from the correctly-merged level into the nested schema by hand.

PR to fix - https://github.com/exoego/rspec-openapi/pull/350

Spence1115:
Fixes #351

## Problem

Heterogeneous array item shapes merge correctly at the level of the array being merged (#267), but when an array-typed property sits *inside* the items being merged and its shape differs **across the outer items**, the nested array is not merged — the schema sticks on the first-encountered shape, and the recorded response example fails to validate against its own schema.

Minimal trigger: `form[0].inputs[].columns[].input` is a text input, `form[1].inputs[].columns[].input` is a dropdown with extra properties. The merged `columns.items.properties.input` schema only contains the text shape. The same shapes merge fine when they appear within a single `columns` array — the bug only triggers when the variation is spread across different outer items.

## Cause

`merge_multi_array_items` (non-recursive mode) has an explicit `'array'` branch that merges item variations, but its twin `merge_multi_recursive` (used inside `build_merged_schema_from_variations`) does not — array-typed properties fall through to `prop_variations.first.dup`, silently dropping later variations. The two methods had drifted.

## Fix

Unify the two paths into a single `merge_multi` that flattens existing `oneOf` entries, recurses into objects **and arrays**, and turns divergent types into `oneOf`. After adding the missing `'array'` branch the two methods were identical except for the `oneOf`-flattening guard, so keeping them separate only invited further drift. The guard is a no-op for the non-recursive path (schemas built by `build_property` never contain `oneOf`), which the regenerated fixtures confirm.

## Testing

- New fixture endpoint `/array_hashes/nested_arrays_across_items` (rails rspec + minitest + hanami) reproducing the cross-item nested shape; golden yaml/json regenerated.
- All doc regeneration diffs are purely additive — no existing fixture output changed.
- Full suite + rubocop pass.

### DIFF
--- a/lib/rspec/openapi/schema_builder.rb
+++ b/lib/rspec/openapi/schema_builder.rb
@@ -377,10 +377,8 @@ class << RSpec::OpenAPI::SchemaBuilder
     merged =
       if prop_variations.size == 1
         prop_variations.first.dup
-      elsif allow_recursive_merge
-        merge_multi_recursive(prop_variations)
       else
-        merge_multi_array_items(prop_variations)
+        merge_multi(prop_variations)
       end
 
     return merged unless merged.is_a?(Hash)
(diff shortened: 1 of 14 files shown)


exoego:
Thank you

PR #356, opened 2026-06-22
fix: preserve inherited openapi summary/operation_id in nested contexts

### LINKED ISSUE #355: `summary` from an outer `describe` is lost (replaced by the Rails action name) when a nested context declares its own `openapi:` metadata
### Summary

When an inner `context`/`describe` sets its own `openapi:` metadata, examples inside it lose the `summary` declared on an ancestor group. The generated operation `summary` then falls back to the Rails action name (`index`, `show`, …), and because multiple examples on the same operation are merged last-wins, that action-name summary can overwrite the real one for the whole operation.

### Reproduction

```ruby
RSpec.describe 'Things', type: :request do
  describe 'GET /things', openapi: { summary: 'List things' } do
    it 'works' do
      get '/things'
      expect(response).to have_http_status(:ok)
    end

    # A nested context that declares ANY openapi metadata of its own
    # replaces the inherited :openapi hash for examples inside it:
    context 'edge case', openapi: { example_mode: :none } do
      it 'also works' do
        get '/things'
        expect(response).to have_http_status(:ok)
      end
    end
  end
end
```

Run with OPENAPI=1.

- Expected: paths./things.get.summary == "List things"
- Actual: paths./things.get.summary == "index" (the action name)

Remove the inner context and the summary is correct — proving the inner openapi: override is what drops it.

This is already latent in this repo's own fixtures: spec/requests/hanami_spec.rb declares describe '#index', openapi: { summary: 'Get a list of tables' } wrapping a context ..., openapi: { example_mode: :multiple }, and the committed golden spec/apps/hanami/doc/openapi.{yaml,json} records the leaked summary: index for GET /tables.

**Root cause**

SharedExtractor.collect_openapi_metadata walks the ancestry by advancing current = current[:parent_example_group] starting from the example metadata. Example metadata has no :parent_example_group key (only group metadata does), so the loop terminates after a single pass and never visits any ancestor group. Once a nested level replaces :openapi, an ancestor's summary/operation_id can no longer be recovered.

summary is the visible symptom because its fallback (route.requirements[:action]) is non-nil and wins the last-wins scalar merge. operation_id leaks identically but is harmless (its nil fallback is compacted out). description had the same class of bug, addressed separately via _fallback_description (v0.26.0).

**Affected versions**

Reproduced on current master (v0.28.0) and v0.26.0; the buggy walk predates it.

K4sku:
Fixes #355

### Problem

When a nested `context`/`describe` declares its own `openapi:` metadata, RSpec *replaces* (does not deep-merge) the inherited `openapi` hash for examples inside it. `SharedExtractor.collect_openapi_metadata` is meant to walk the example-group ancestry and re-merge ancestor levels, but it advanced via `current[:parent_example_group]` starting from the **example** metadata — which has no `:parent_example_group` key (only *group* metadata does). The loop therefore ran once and never visited any ancestor group, so an outer `summary`/`operation_id` was dropped.

For `summary` this is user-visible: the dropped value falls back to the Rails action name (`route.requirements[:action]`), and because records for the same operation are merged last-wins, that action-name summary can overwrite the real summary declared on an outer `describe`.

### Fix

Walk the group ancestry the same way `RSpec::Core::Metadata.ascending` does: start from the example's own `:openapi`, then `fetch(:example_group) { metadata[:parent_example_group] }` and follow `:parent_example_group` upward, unshifting so the nearest level still wins per key.

Using `fetch` with a fallback block mirrors rspec-core exactly: it makes the walk correct whether handed example or group metadata, and avoids tripping the deprecated `:example_group`-on-group-metadata default proc.

### Tests

- Unit specs in `spec/rspec/shared_extractor_spec.rb`.
- Corrects the pre-existing Hanami golden (`spec/apps/hanami/doc/openapi.{yaml,json}`): `GET /tables` summary changes from the leaked `index` to the declared `Get a list of tables`. This is the only fixture change — the Rails fixtures don't use the pattern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,10 +190,16 @@ RSpec::OpenAPI.description_builder = -> (example) { example.description }
 
 # Generate a custom summary, given an RSpec example
 # This example uses the summary from the example_group.
+# NOTE: a fixed `dig` only reaches one specific depth. `summary` declared on an
+# ancestor group is already inherited automatically (see "Inheritance of
+# `openapi:` metadata"), so a builder is only needed for fully custom logic.
 RSpec::OpenAPI.summary_builder = ->(example) { example.metadata.dig(:example_group, :openapi, :summary) }
 
 # Generate a custom tags, given an RSpec example
-# This example uses the tags from the parent_example_group
+# This example uses the tags from the parent_example_group.
+# NOTE: `dig(:example_group, :parent_example_group, ...)` only matches tags
+# declared exactly two levels up; it won't find tags on deeper ancestors. Rely
+# on the built-in inheritance unless you need custom resolution.
 RSpec::OpenAPI.tags_builder = -> (example) { example.metadata.dig(:example_group, :parent_example_group, :openapi, :tags) }
 
 # Configure custom format for specific properties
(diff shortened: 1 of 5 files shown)


exoego:
@K4sku Thanks.
The logic change looks good to me.
I've updated docs, test and PR title.

PR #370, opened 2026-06-26
Fix first-request capture for multi-call examples

### LINKED ISSUE #371: OpenAPI metadata gets mixed up when performing multiple requests on the same URL in an example
I noticed that, when in an example the same URL is called twice, there is a mixup of metadata on endpoints in the OpenAPI spec. In my cas I noticed that the summary of the delete call appeared as example text for the get example returning 404.

The pattern is this:

```ruby
describe 'GET /example', openapi: ( summary: 'get example' } do
  it 'succeeds' {}
  it 'not found' {}
end

describe 'DELETE /example', openapi: {summary: 'delete example' } do
  it 'deletes an example' do
    delete '/example'
    [...]
    get '/example'
    expect(last_response.status).to eq(404)
  end
```
That will most likely end up in having the get URL error 404 example being labelled as 'deletes an example' instead of 'not found'.

The proposed fix in #370 uses a RequestRecorder that captures the first call and having the request_response return that one over the one derived from the context. I also added a few test cases to illustrate the issue and the fix.

Kris-LIBIS:
Closes #371

This fixes OpenAPI generation for examples that make multiple requests against the same path.

Problem:
- When an example performs a follow-up request after the request that should define the documented operation, the generator could reuse the later request/response metadata and contaminate the operation details for the shared path.

Fix:
- Capture the first request/response exchange seen in an example and prefer that exchange during extraction.
- This keeps operation metadata stable when a single example exercises multiple verbs or repeated calls on the same path.

Tests:
- Added a regression spec that reproduces the multi-request shared-path case and verifies the generated OpenAPI keeps the correct GET/DELETE metadata separate.
- Verified locally with the new regression spec passing.


exoego:
Thanks for the PR.

But I don't like breaking the current contract (last request wins).
Instead, I'd like to let users to specify which request to be documented (and defaults to last request wins).
Feel free to open a new issue if you find the new feature doesn't satisfy your usecases. 
https://github.com/exoego/rspec-openapi/pull/376

PR #381, opened 2026-07-10
fix: cast boolean query parameters instead of typing them as string

### LINKED ISSUE #380: bug: boolean query parameters are typed as `string`, though integer query params are cast
One of our apps has an endpoint with boolean flags in the query string, e.g. `GET /recent_solutions?show_unlaunchable_solutions=false&use_zappi_form_actions=true`. Every one of those flags is typed as `string` in the generated schema:

```yaml
- name: use_zappi_form_actions
  in: query
  required: false
  schema:
    type: string
  example: "true"
```

when the parameter is really a boolean and should be:

```yaml
- name: use_zappi_form_actions
  in: query
  required: false
  schema:
    type: boolean
  example: true
```

I know query strings are untyped on the wire, so at first glance `string` looks defensible. But the gem already doesn't treat query params as opaque strings — `SchemaBuilder#try_cast` recovers integers:

```ruby
# lib/rspec/openapi/schema_builder.rb
def try_cast(value)
  Integer(value)
rescue TypeError, ArgumentError
  value
end
```

So `?page=1` correctly generates `type: integer, example: 1` — the existing flat-query test in `spec/requests/rails_spec.rb` passes `page: '1'` (a string) and asserts `type: integer` in the fixture. Booleans are the same case: a value that arrives as a string but has an obvious richer type. `try_cast` just never had a branch for them, so the outcome is inconsistent — integer flags get their real type, boolean flags don't.

Downstream this means our type generator (Orval) emits `string` for parameters the API treats as booleans, so callers end up passing `"true"` / `"false"` string literals instead of a boolean.

The fix is the missing sibling branch in `try_cast`:

```ruby
def try_cast(value)
  return true if value == 'true'
  return false if value == 'false'

  Integer(value)
rescue TypeError, ArgumentError
  value
end
```

This carries the same small, pre-existing trade-off the integer cast already makes: a param whose genuine value is the literal string `"1"` is cast to an integer today, and a param whose genuine value is literally `"true"` / `"false"` would now be cast to a boolean. In practice a query param that reads `true` / `false` is almost always a boolean flag, so I think the consistency is worth it — but happy to gate it behind an option if you'd rather keep it opt-in.

PR to fix - https://github.com/exoego/rspec-openapi/pull/381

Spence1115:
Fixes #380

## Problem

Boolean flags in a query string are typed as `string` in the generated schema. Given `GET /tables?active=true`, the parameter is recorded as `type: string, example: "true"` when it is really a boolean (`type: boolean, example: true`). Downstream, type generators like Orval then emit `string` for parameters the API treats as booleans, so callers pass `"true"` / `"false"` string literals instead of a boolean.

## Cause

Query params arrive as strings, but the gem does not treat them as opaque — `SchemaBuilder#try_cast` already recovers integers, so `?page=1` correctly generates `type: integer`. `try_cast` simply never had a branch for booleans, leaving the behaviour inconsistent: integer flags get their real type, boolean flags do not.

## Fix

Add the missing sibling branch to `try_cast`, casting the literal `"true"` / `"false"` to a boolean before the integer attempt:

```ruby
def try_cast(value)
  return true if value == 'true'
  return false if value == 'false'

  Integer(value)
rescue TypeError, ArgumentError
  value
end
```

This carries the same small, pre-existing trade-off the integer cast already makes — a param whose genuine value is the literal string `"1"` is cast to an integer today, and one whose value is literally `"true"` / `"false"` is now cast to a boolean. A query param reading `true` / `false` is virtually always a boolean flag. Happy to gate it behind an option if you would rather keep it opt-in.

## Testing

- Added a boolean query param (`active: 'true'`) to the existing rails flat-query spec; it now generates `type: boolean`. Golden yaml/json regenerated across 3.0 / 3.1 / 3.2.
- The smart-merge suite already sends `show_columns: true` to `/tables`, so its `expected.yaml` fixture updated too — a second, independent confirmation.
- All doc regeneration diffs are purely additive apart from the two `show_columns` type/example lines. hanami and minitest fixtures are unchanged (neither passes boolean-shaped query params).
- Full suite passes (63 examples, 0 failures).

### DIFF
--- a/lib/rspec/openapi/schema_builder.rb
+++ b/lib/rspec/openapi/schema_builder.rb
@@ -292,6 +292,9 @@ class << RSpec::OpenAPI::SchemaBuilder
 
   # Convert an always-String param to an appropriate type
   def try_cast(value)
+    return true if value == 'true'
+    return false if value == 'false'
+
     Integer(value)
   rescue TypeError, ArgumentError
     value
--- a/spec/apps/rails/doc/rspec_openapi.json
+++ b/spec/apps/rails/doc/rspec_openapi.json
@@ -3404,6 +3404,15 @@
             "example": true
           },
           {
+            "name": "archived",
+            "in": "query",
+            "required": false,
+            "schema": {
+              "type": "boolean"
+            },
+            "example": false
+          },
+          {
             "name": "filter[name]",
             "in": "query",
             "required": false,
--- a/spec/apps/rails/doc/rspec_openapi.json
+++ b/spec/apps/rails/doc/rspec_openapi.json
@@ -3395,6 +3395,15 @@
             "example": "token"
           },
           {
+            "name": "active",
+            "in": "query",
+            "required": false,
+            "schema": {
+              "type": "boolean"
+            },
+            "example": true
+          },
+          {
             "name": "filter[name]",
             "in": "query",
             "required": false,
--- a/spec/apps/rails/doc/rspec_openapi.yaml
+++ b/spec/apps/rails/doc/rspec_openapi.yaml
@@ -2086,6 +2086,12 @@ paths:
         schema:
           type: string
         example: token
+      - name: active
+        in: query
+        required: false
+        schema:
+          type: boolean
+        example: true
       - name: filter[name]
         in: query
         required: false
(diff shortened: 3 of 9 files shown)


exoego (approved):
LGTM

PR #383, opened 2026-07-15
fix: do not treat an empty security array as a security requirement

### LINKED ISSUE #382: [BUG] cleanup_conflicting_security_parameters! raises NoMethodError on an operation with security: [] and parameters
## Summary

An operation marked as public with `security: []` **and** carrying any parameter (path or query) crashes schema generation:

```
NoMethodError: undefined method `keys' for nil
  lib/rspec/openapi/schema_cleaner.rb:106:in `remove_parameters_conflicting_with_security_scheme!'
  lib/rspec/openapi/schema_cleaner.rb:59:in `block (2 levels) in cleanup_conflicting_security_parameters!'
  lib/rspec/openapi/result_recorder.rb:57:in `cleanup_schema!'
```

It raises inside an `after(:suite)` hook, so the whole recording run exits non-zero which breaks CI setups that regenerate the schema and check it with `git diff --exit-code`.

Reproduced on **0.28.0**; the faulty code is unchanged in **0.30.0**.

## Cause

`schema_cleaner.rb:104-106`:

```ruby
return unless path_definition[:security] # [] is truthy in Ruby -> passes
return unless path_definition[:parameters] # needs a parameter
return unless path_definition.dig(:security, 0).keys.include?(...) # dig -> nil -> NoMethodError
```

The first guard reads as "no security, nothing to do", but an empty array is truthy in Ruby, so it lets `security: []` through. `dig(:security, 0)` then returns `nil` and `.keys` raises.

`security: []` is the standard OpenAPI way to opt an operation out of a document-level security requirement, and it is a first-class supported input here: `security` is a documented metadata param (`record.rb:20` - `@param [Array]`), read in `shared_extractor.rb:30` and written straight to the schema in `schema_builder.rb:29`.

This looks like an edge case missed when #160 was fixed by #166: the cleanup assumes `security` is non-empty whenever the key is present.

## Reproduction

Requires all three: `security_schemes` configured, an operation with `security: []`, and a parameter on it. That combination is probably why it went unnoticed. a public endpoint with no parameter, or an authenticated endpoint with one, both work fine.

Minimal, no RSpec needed:

```ruby
require "rspec/openapi"

spec = {
  paths: {
    :"/widgets/{id}" => {
      get: {
        security: [],
        parameters: [{ name: "id", in: "path", required: true, schema: { type: "string" } }],
        responses: { "200" => { description: "ok" } }
      }
    }
  },
  components: {
    securitySchemes: { "bearerAuth" => { type: "http", scheme: "bearer" } }
  }
}

RSpec::OpenAPI::SchemaCleaner.cleanup_conflicting_security_parameters!(spec)
# => NoMethodError: undefined method `keys' for nil
```

Controls, both fine:

- same operation with `security: [{ bearerAuth: [] }]` -> no crash
- same operation public but with `parameters` removed -> no crash

In a real suite this is reached via:

```ruby
RSpec::OpenAPI.security_schemes = {
  "bearerAuth" => { type: "http", scheme: "bearer", bearerFormat: "JWT" }
}

# GET /merchants/:id/catalog - public, so it opts out of the document-level default
it "returns the catalog", openapi: { security: [] } do
  get "/api/v1/merchants/#{merchant.id}/catalog"
  expect(response).to have_http_status(:ok)
end
```

## Suggested fix

Make the first guard mean what it says, an operation with no security requirement has no scheme to conflict with, so there is nothing to strip:

```ruby
return if path_definition[:security].blank? # or: .nil? || .empty?
```

Happy to send a PR with a regression spec if that direction works for you.

## Environment

- rspec-openapi 0.28.0 (checked 0.30.0: same code)
- Ruby 3.x / Rails 8 API-only

Dashby7:
Fixes #382.

An operation opted out of the document-level requirement with `security: []` has
no scheme to conflict with, but `[]` is truthy, so the guard let it through and
`dig(:security, 0)` returned nil.

The added spec fails on master with the reported NoMethodError. The three other
examples cover the existing behaviour and pass either way.


exoego:
Thank you.
Replaced the unit test with integration test

PR #397, opened 2026-07-28
Unwrap Rails' _json root from recorded JSON request bodies

carlosantoniodasilva:
Rails' default JSON parser wraps a request body that isn't a Hash - a
top-level array, string, number, and so on - as `{ _json: <body> }`
(ActionDispatch::Http::Parameters::DEFAULT_PARSERS). Because the recorder
reads `request.request_parameters` verbatim, that internal wrapper leaked
into the generated document: a `POST` with a bare-array body was documented
as an object with a required `_json` array property, and the example was
`{ _json: [...] }` - a shape no client ever sends.

Unwrap the wrapper at the source, next to the existing ParamsWrapper
workaround, so both the schema and the example are fixed in one place. The
unwrap applies only when `_json` is the sole key, so an object that
genuinely carries a `_json` key alongside others is left untouched.

`build_example` assumed a Hash root and called `adjust_params`, which is
`transform_values!`; it now dispatches through `adjust_value` so an array
or scalar root works.

https://github.com/rails/rails/blob/fa8f0812160665bff083a089d2bb2fc1817ea03e/actionpack/lib/action_dispatch/http/parameters.rb#L12-L17

### DIFF
--- a/README.md
+++ b/README.md
@@ -897,7 +897,7 @@ responses:
   '400': { ... }
 ```
 
-Mixing `:single` (some tests) with `:multiple` (others) on the same endpoint also works for `requestBody` - 
+Mixing `:single` (some tests) with `:multiple` (others) on the same endpoint also works for `requestBody` -
 the merger up-converts the singular `example:` into the `examples:` map automatically
 (see [Merge Behavior with Mixed Modes](#merge-behavior-with-mixed-modes) below).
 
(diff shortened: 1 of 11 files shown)


exoego:
Thanks.
I have verified this with https://github.com/ohler55/oj and it worked too.
Oj just replaces `ActiveSupport::JSON.decode` seemingly, so Raiils with Oj till wraps with `{ _data: ... }`.

PR #454, opened 2026-09-18
Keep a `$ref` nested inside an inline envelope alive

carlosantoniodasilva:
`ComponentsUpdater` only keeps a nested `$ref` schema alive across regens when it's reachable one of two ways: as the top-level body of a request/response, or nested inside a schema that's *already* a component (`find_non_top_level_nested_refs`, which walks `components.schemas.*`).

A `$ref` sitting directly inside an inline object — one that was never itself promoted to a component — falls outside both searches. It gets silently dropped from `components/schemas` on the next regen, even though the `$ref` pointing at it is still there in the spec.

Example — this is stable under repeated regens today:

```yaml
paths:
  /rooms:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/RoomList'
components:
  schemas:
    RoomList:
      type: object
      properties:
        rooms:
          type: array
          items:
            $ref: '#/components/schemas/Room'
    Room:
      type: object
      properties:
        id: { type: integer }
        name: { type: string }
```

...but drop the `RoomList` wrapper and inline it directly:

```yaml
paths:
  /rooms:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                type: object
                properties:
                  rooms:
                    type: array
                    items:
                      $ref: '#/components/schemas/Room'
components:
  schemas:
    Room:
      type: object
      properties:
        id: { type: integer }
        name: { type: string }
```

...and the next `OPENAPI=1 rspec` run deletes `Room` from `components/schemas` entirely. Nothing about this response is a top-level `$ref` (it's a plain object with an array property, not a bare `$ref/array-of-$ref/oneOf`), and `Room` isn't nested inside some other already-registered component either — so it's invisible to both of `ComponentsUpdater`'s existing passes.

This adds a new search for nested `$ref`'s under `paths.*.*.{requestBody,responses}.*.content.application/json.schema`
instead of just `components.schemas`, so it is able to find and keep those inlined `$ref`s as well.

It also refactors the updater by extracting methods to apply each search (components/inline), and shares some pieces between them.


spec/rspec/components_updater_spec.rb
```diff
@@ -0,0 +1,157 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe 'components updater spec' do
```
carlosantoniodasilva, line 5: I ended up creating a separate unit-type spec for the last bugfix, since it was easier to create a scenario that does not have a top level $ref that way. Happy to try something else.

I also considered making these "let" variables `YAML.safe_load` to shorten them a bit, but kept as Hash objects since that seems to be the common way around other specs. Let me know if it'd be preferable.

<details><summary>Example</summary>
<p>

```diff
diff --git a/spec/rspec/components_updater_spec.rb b/spec/rspec/components_updater_spec.rb
index 8c871cc..573a40f 100644
--- a/spec/rspec/components_updater_spec.rb
+++ b/spec/rspec/components_updater_spec.rb
@@ -1,79 +1,59 @@
 # frozen_string_literal: true
 
 require 'spec_helper'
+require 'yaml'
 
 RSpec.describe 'components updater spec' do
   describe 'a $ref nested inline, with no top-level $ref anywhere in the document' do
     let(:base) do
-      {
-        paths: {
-          '/rooms': {
-            get: {
-              responses: {
-                '200': {
-                  content: {
-                    'application/json': {
-                      schema: {
-                        type: 'object',
-                        properties: {
-                          room: {
-                            '$ref': '#/components/schemas/Room',
-                          },
-                        },
-                      },
-                    },
-                  },
-                },
-              },
-            },
-          },
-        },
-        components: {
-          schemas: {
-            Room: {
-              type: 'object',
-              description: 'should be preserved',
-              properties: {
-                id: { type: 'integer' },
-                stale_field: { type: 'string' },
-              },
-            },
-          },
-        },
-      }
+      YAML.safe_load(<<~YAML, symbolize_names: true)
+        paths:
+          /rooms:
+            get:
+              responses:
+                '200':
+                  content:
+                    application/json:
+                      schema:
+                        type: object
+                        properties:
+                          room:
+                            $ref: '#/components/schemas/Room'
+        components:
+          schemas:
+            Room:
+              type: object
+              description: should be preserved
+              properties:
+                id:
+                  type: integer
+                stale_field:
+                  type: string
+      YAML
     end
 
     let(:fresh) do
-      {
-        paths: {
-          '/rooms': {
-            get: {
-              responses: {
-                '200': {
-                  content: {
-                    'application/json': {
-                      schema: {
-                        type: 'object',
-                        properties: {
-                          room: {
-                            type: 'object',
-                            properties: {
-                              id: { type: 'integer' },
-                              name: { type: 'string' },
-                            },
-                            required: ['id', 'name'],
-                          },
-                        },
-                        required: ['room'],
-                      },
-                    },
-                  },
-                },
-              },
-            },
-          },
-        },
-      }
+      YAML.safe_load(<<~YAML, symbolize_names: true)
+        paths:
+          /rooms:
+            get:
+              responses:
+                '200':
+                  content:
+                    application/json:
+                      schema:
+                        type: object
+                        properties:
+                          room:
+                            type: object
+                            properties:
+                              id:
+                                type: integer
+                              name:
+                                type: string
+                            required: [id, name]
+                        required: [room]
+      YAML
     end
 
     it 'still refreshes the schema instead of dropping it' do
@@ -88,64 +68,45 @@ RSpec.describe 'components updater spec' do
 
   describe 'no $ref anywhere in the document' do
     let(:base) do
-      {
-        paths: {
-          '/ping': {
-            get: {
-              responses: {
-                '200': {
-                  content: {
-                    'application/json': {
-                      schema: {
-                        type: 'object',
-                        properties: {
-                          ok: { type: 'boolean' },
-                        },
-                      },
-                    },
-                  },
-                },
-              },
-            },
-          },
-        },
-        components: {
-          schemas: {
-            Unrelated: {
-              type: 'object',
-              properties: {
-                id: { type: 'integer' },
-              },
-            },
-          },
-        },
-      }
+      YAML.safe_load(<<~YAML, symbolize_names: true)
+        paths:
+          /ping:
+            get:
+              responses:
+                '200':
+                  content:
+                    application/json:
+                      schema:
+                        type: object
+                        properties:
+                          ok:
+                            type: boolean
+        components:
+          schemas:
+            Unrelated:
+              type: object
+              properties:
+                id:
+                  type: integer
+      YAML
     end
 
     let(:fresh) do
-      {
-        paths: {
-          '/ping': {
-            get: {
-              responses: {
-                '200': {
-                  content: {
-                    'application/json': {
-                      schema: {
-                        type: 'object',
-                        properties: {
-                          ok: { type: 'boolean' },
-                        },
-                        required: ['ok'],
-                      },
-                    },
-                  },
-                },
-              },
-            },
-          },
-        },
-      }
+      YAML.safe_load(<<~YAML, symbolize_names: true)
+        paths:
+          /ping:
+            get:
+              responses:
+                '200':
+                  content:
+                    application/json:
+                      schema:
+                        type: object
+                        properties:
+                          ok:
+                            type: boolean
+                        required: [ok]
+      YAML
     end
 
     it 'leaves other, unrelated components alone rather than wiping them' do
```

</p>
</details>

PR #456, opened 2026-09-20
Fix NullableConverter nesting type arrays on round-trip

### LINKED ISSUE #451: NullableConverter produces an invalid nested type array for multi-type nullable schemas
### Summary

When an OpenAPI 3.1+ schema contains multiple non-null types and `null`, `NullableConverter` does not round-trip it correctly.

A valid schema such as:

```yaml
type: [string, number, boolean, object, array, "null"]
```

is rewritten as:

```yaml
type:
  - [string, number, boolean, object, array]
  - "null"
```

The nested array is invalid JSON Schema and causes consumers such as `oasdiff` to reject the generated document.

### Reproduction

```ruby
require "rspec/openapi"

schema = {
  type: ["string", "number", "boolean", "object", "array", "null"],
}

RSpec::OpenAPI::NullableConverter.normalize!(schema)
RSpec::OpenAPI::NullableConverter.to_json_schema!(schema)

pp schema
```

### Actual result

```ruby
{
  type: [
    ["string", "number", "boolean", "object", "array"],
    "null",
  ],
}
```

### Expected result

```ruby
{
  type: ["string", "number", "boolean", "object", "array", "null"],
}
```

The conversion should preserve a flat array of type names.

### Cause
`nullable_to_type_null!` always wraps the existing type:
```ruby
[type, "null"]
```

After `normalize!`, `type` can already be an array containing multiple non-null types. Wrapping that array creates the invalid nested result.

The array case could append `null` instead:

```ruby
case type
when nil
  "null"
when Array
  type | ["null"]
else
  [type, "null"]
end
```

### Impact
`oasdiff validate` and `oasdiff breaking` fail while loading the generated contract:

```text
json: cannot unmarshal array into field Schema.type of type string
```

This also prevents tools expecting a valid JSON Schema `type` array from consuming the generated OpenAPI document.

### Workaround

Expressing the union with `anyOf` avoids the faulty conversion:

```yaml
anyOf:
  - type: string
  - type: number
  - type: boolean
  - type: object
  - type: array
  - type: "null"
```

### Environment

- `rspec-openapi`: 0.33.0
- The same converter implementation is present in 0.33.1
- OpenAPI: 3.2.0
- Ruby: 4.0.5

niukanen1:
When a schema's `type` is an array of several non-null types and the schema is nullable, `NullableConverter#to_json_schema!` wrapped the whole array in another array together with `"null"`, producing an invalid nested type like `[[string, number, boolean, object, array], "null"]`. `normalize!` leaves such schemas with a flat multi-type array plus `nullable: true`, so the wrap hits every 3.1+ round-trip of a multi-type nullable schema.

Append `"null"` to the existing array instead of nesting it (using `|` so an already-present `"null"` is not duplicated). The round-trip of `type: [string, number, boolean, object, array, "null"]` now stays flat, which is what JSON Schema consumers such as `oasdiff` expect.

Fixes #451


exoego:
@niukanen1 Thanks for the PR.

In this repo, integration tests (running specs on Rails and generate YAML/JSON) are preferred instead of unit tests.
Because unit test is tightly bound to concrete classes/methods, which are likely replaced/removed by refactoring as time goes.
You can get inspiration from the existing integration tests and add a tiny test for this.

niukanen1:
Thanks for the direction — replaced the unit test with an integration spec in 2b15fd1, following the existing hand-edited round-trip pattern:

- a request spec re-records a 200 over a document whose hand-written 409 response carries a multi-type nullable schema (`type: [string, integer, "null"]`)
- the spec asserts the regenerated document keeps the type array flat instead of nesting it
- the seed document is restored afterwards, like the other round-trip specs

Verified the spec fails against the nested-array form and passes with the fix; full suite (89 examples) and rubocop are clean

